PR #1, opened 2021-09-20
feat: stabilize Ingress for cdk8s-plus-22

Chriscbr:
Ingress v1beta1 is no longer available in k8s 1.22.0, so we are now using the stabilized Ingress v1 API.

The kubernetes [deprecation guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122) lists most of the differences/changes that have been made. Most of the changes are just renaming properties so I did not make significant changes to the API in cdk8s+, but there is a newly required field named "pathType" for all Ingress paths, which can take on the values of "Prefix", "Exact", and "ImplementationSpecific".

The migration guide states that "To match the undefined v1beta1 behavior, use ImplementationSpecific". I added "Prefix" since its behavior is most clearly documented for users and I think it makes sense as the better default between "Prefix" and "Exact". But I think there's also a decent argument for providing "ImplementationSpecific" as the default so that the output manifest behaves identically between cd8s-plus-21 and cdk8s-plus-20, so I'm open to changing this.

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -2,12 +2,25 @@ const child = require('child_process');
 const path = require('path');
 const { JsiiProject } = require('projen');
 
+const DEFAULT_K8S_VERSION = '22';
 const SPEC_VERSION = k8sVersion();
 const K8S_VERSION = `1.${SPEC_VERSION}.0`;
 
 function k8sVersion() {
   const branch = child.execSync('git rev-parse --abbrev-ref HEAD').toString().trim();
-  return branch.replace('k8s.', '');
+  if (branch.startsWith('k8s.') && branch.length >= 6) {
+    return branch.substring(4, 6);
+  } else {
+    // if the branch name doesn't start with k8s.XX, we're probably running on a fork
+    // so assume that we are building for the latest version, i.e. k8s.22
+    console.log(
+      'Warning: The current branch name doesn\'t start with "k8s.XX", so we ' +
+      `are defaulting to built for k8s 1.${DEFAULT_K8S_VERSION}.0. If you ` +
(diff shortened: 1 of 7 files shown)


src/ingress.ts
```diff
@@ -152,7 +176,8 @@ export class IngressV1Beta1 extends Resource {
 
       const host = rule.host ?? '';
       const backend = rule.backend;
-      const path = rule.path;
+      const path = rule.path ?? '/';
+      const pathType = rule.pathType ?? HttpIngressPathType.PREFIX;
```
Chriscbr, line 180: In the raw spec, "path" is marked as an optional field -- but it is required if the pathType is either "Path" or "Prefix". Our `addDefaultBackend` and `addHostDefaultBackend` APIs don't require the user to provide any path, but I think "/" is the appropriate default here.

.projenrc.js
```diff
@@ -2,12 +2,25 @@ const child = require('child_process');
...
 const { JsiiProject } = require('projen');
 
+const DEFAULT_K8S_VERSION = '22';
 const SPEC_VERSION = k8sVersion();
 const K8S_VERSION = `1.${SPEC_VERSION}.0`;
 
 function k8sVersion() {
   const branch = child.execSync('git rev-parse --abbrev-ref HEAD').toString().trim();
-  return branch.replace('k8s.', '');
+  if (branch.startsWith('k8s.') && branch.length >= 6) {
```
Chriscbr, line 11: We need to specify what happens if the branch is not one of "k8s.22", "k8s.21", or "k8s.20" otherwise 'npx projen' will generate unexpected outputs and the code will not build.

PR #20, opened 2021-09-24
chore: use caret for cdk8s peer dependency

Chriscbr:
Using a caret for our peer dependency better models the compatible versions of the cdk8s core library.

In addition, remove some config from .projenrc.js that set pinned dev dependencies to false because I believe this was is an artifact from when the all of the cdk8s packages were a monorepo that no longer applies.


Chriscbr:
@iliapolo I'm going to merge this one in since I think it's a low risk change - if it doesn't work, I'll revert it.

PR #37, opened 2021-10-04
chore: DCO

iliapolo:
Part of https://github.com/cdk8s-team/cdk8s/issues/508


eladb (approved):
Feels like this could be a useful projen feature...

iliapolo:
> Feels like this could be a useful projen feature...

Yeah - there are just a few subtleties here that made me add this directly at the moment. 
But for sure an interesting use-case.

PR #51, opened 2021-10-05
feat: expose service via ingress

iliapolo:
I was looking at our canonical [example](https://cdk8s.io/docs/latest/plus/ingress/) for using `Ingress` with CDK8s+ and felt we could simplify this even more. 

This PR introduces an `exposeViaIngress` method on a **service**, which will create the necessary `IngressBackend` and the corresponding rule on the `Ingress`. 

In addition, add an `exposeViaIngress` method on a **deployment**, which will first expose it via a service, and then expose the service via an `Ingress`. 

Following this PR, our example can be simplified to:

```ts
new kplus.Deployment(this, text, {
  containers: [
    {
      image: 'hashicorp/http-echo',
      args: [ '-text', 'hello ingress' ],
      port: 5678
    }
  ]
}).exposeViaIngress('/hello');
```

BREAKING CHANGE: `deployment.expose` renamed to `deployment.exposeViaService`

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -304,7 +304,7 @@
         "RELEASE": "true",
         "MAJOR": "1",
         "PRERELEASE": "beta",
-        "PREFIX": "cdk8s-plus-22/"
+        "RELEASE_TAG_PREFIX": "cdk8s-plus-22/"
       },
       "steps": [
         {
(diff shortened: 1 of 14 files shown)


eladb (approved):
Nice. I like the names

PR #211, opened 2021-12-15
feat: Secret volume support

### LINKED ISSUE #34: Secret volume support
I could be missing something but don't see a way to use secret volumes with this library.

The `Volume` class has a protected constructor and only supports config maps and empty directories.

The `Deployment` and `Container` classes require modifications to pass through methods that only accept the `Volume` class.

Is this not supported right now? 

As a side note it would be useful to have some advanced access to cover areas not implemented by the high level abstractions, something like the following for this use case: 
```typescript
Volume.create({ name: "secret", secret: { secretName: "secret-name" } } as k8.Volume) // create volume from raw kubernetes spec
```

ambasta:
Fixes #34

### DIFF
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -86,6 +86,21 @@ export class Volume {
     });
   }
 
+  /**
+   * Populate the volume from a Secret.
+   *
+   * A secret volume is used to pass sensitive information, such as passwords, to Pods.
+   * You can store secrets in the Kubernetes API and mount them as files for use by pods
+   * without coupling to Kubernetes directly.
+   *
+   * secret volumes are backed by tmpfs (a RAM-backed filesystem)
+   * so they are never written to non-volatile storage.
+   *
+   * @see https://kubernetes.io/docs/concepts/storage/volumes/#secret
+   *
+   * @param secret The secret to use to populate the volume.
+   * @param options Options
+   */
   public static fromSecret(secret: ISecret, options: SecretVolumeOptions = { }): Volume {
     return new Volume(options.name ?? `secret-${secret.name}`, {
       secret: {
--- a/src/volume.ts
+++ b/src/volume.ts
@@ -1,6 +1,7 @@
 import { Size } from 'cdk8s';
 import { IConfigMap } from './config-map';
 import * as k8s from './imports/k8s';
+import { ISecret } from './secret';
 
 /**
  * Volume represents a named volume in a pod that may be accessed by any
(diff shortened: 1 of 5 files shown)


src/volume.ts
```diff
@@ -98,6 +99,30 @@ export class Volume {
     });
   }
 
+  public static fromSecret(secret: ISecret, options: SecretVolumeOptions = { }): Volume {
```
Chriscbr, line 104: Please add unit tests for this function in `volume.test.ts`

src/volume.ts
```diff
@@ -220,3 +245,44 @@ export enum EmptyDirMedium {
    */
   MEMORY = 'Memory'
 }
+
+export interface SecretVolumeOptions {
```
Chriscbr, line 258: Please add a docstring to this interface.

src/volume.ts
```diff
@@ -98,6 +99,30 @@ export class Volume {
...
+      const result = new Array<k8s.KeyToPath>();
+      for (const key of Object.keys(items).sort()) {
+        result.push({
+          key,
+          path: items[key].path,
+          mode: items[key].mode,
+        });
+      }
+      return result;
+    }
```
Chriscbr: It looks like this function is the same as the one in `fromConfigMap` - can we move this so there's just one copy that's reused in both places?

Chriscbr (changes requested):
Thanks for the contribution @ambasta ! Just a few comments before we get this merged in.

src/volume.ts
ambasta: Done

src/volume.ts
ambasta: Added tests in test/volume.test.ts

src/volume.ts
ambasta: Created an internal private static function which both Volume.fromSecret and Volume.fromConfigMap. Let me know if I should move this to utils or a standalone function instead. Not sure if the function should be documented since it is not meant for public usage.

src/volume.ts
Chriscbr: @ambasta Sorry I didn't catch this earlier - but can you also add a docstring to this `fromSecret` function? See `fromConfigMap` as a reference

Chriscbr (changes requested):

src/volume.ts
ambasta: I am surprised I missed this as well. I was sure I'd added this before, but pushed an updated commit to address this. Let me know if I should squash the two commits into one.

Chriscbr (approved):

Chriscbr:
Great work @ambasta !

PR #263, opened 2022-01-13
feat: Env var from fieldRef/resource support

ambasta:
Initial work to add EnvVars from fieldRef and resourceFieldRef.

### ToDo: 

- [x] ~~Write test cases.~~

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -4,6 +4,90 @@ import { Probe } from './probe';
 import { SecretValue } from './secret';
 import { Volume } from './volume';
 
+export enum EnvFieldPaths {
+  /**
+   * The name of the pod.
+   */
+  POD_NAME = 'metadata.name',
+
+  /**
+   * The namespace of the pod.
+   */
+  POD_NAMESPACE = 'metadata.namespace',
+
+  /**
+   * The uid of the pod.
+   */
+  POD_UID = 'metadata.uid',
+
+  /**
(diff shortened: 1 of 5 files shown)


src/container.ts
```diff
@@ -97,6 +228,52 @@ export class EnvValue {
...
   }
 
+  /**
+   *
+   * Create a value from a field reference.
+   *
+   * @param fieldPath: The field reference.
+   * @param options: Additional options.
+   */
+  public static fromFieldRef(fieldPath: EnvFieldPaths | FieldPaths, options: EnvValueFromFieldRefOptions = {}): EnvValue {
```
Chriscbr: We generally avoid using type unions in APIs if possible because when cdk8s-plus gets transpiled by JSII to other languages like Java and .NET, users in those languages will not see any typing information (because those languages don't support type unions natively). See https://aws.github.io/jsii/specification/2-type-system/#type-unions for more info.

Perhaps we could combine all of the allowed values into `EnvFieldPaths`? I'm okay with having some redundancy among enums if needed.

Chriscbr (changes requested):
Nice work! Just one comment about the API

src/container.ts
ambasta: Hi, have currently merged the two into a EnvFieldPaths.

Chriscbr (approved):

PR #275, opened 2022-01-17
feat: add PersistentVolumeClaim support

### LINKED ISSUE #176: Construct for PersistentVolumeContainer (PVC)
There does not seem to be a way to attach a PVC to a container in a  deployment with cdk8s+.  My goal is to persist data added to a DB. I know I could go with a stateful set, but that is beyond what my simple sample use case requires. Basically, I want to generate something like this (partial manifest shown below):

```yaml
       ..... 
         volumeMounts:
        - mountPath: /var/lib/mysql
          name: mariadb-claim0
      volumes:
      - name: mariadb-claim0
        persistentVolumeClaim:
          claimName: mariadb-claim0
     .....
```

In theory, I can define the volume_mount that contains a Volume and a PVC:

```python
import cdk8s_plus_22 as kplus
from imports import k8s

mariadb_pvc = k8s.KubePersistentVolumeClaim(mariaDB_chart,
                                   'mariadbPvc',
                                   spec=k8s.PersistentVolumeClaimSpec(access_modes=['ReadWriteOnce'],
                                                        resources=k8s.ResourceRequirements(
                                                         requests={
                                                                  "storage": k8s.Quantity.from_string(
                                                                  "100Mi")})))

vm = kplus.VolumeMount(path="/var/lib/mysql", volume=k8s.Volume(
                        name="test",
                        persistent_volume_claim=k8s.PersistentVolumeClaimVolumeSource(claim_name="mariadbPvc")
                       ))
```

I then try to add the volume mount to the container:

```python
kplus.ContainerProps(image='mariadb',
           name="mariadbcont",
           port=3306,
           volume_mounts=[vm]
          )
```

But  I get the following error:

```
raise JSIIError(resp.error) from JavaScriptError(resp.stack)
jsii.errors.JSIIError: Expected object reference, got {"$jsii.struct":{"fqn":"k8s.Volume","data":{"name":"test","awsElasticBlockStore":null,"azureDisk":null,"azureFile":null,"cephfs":null,"cinder":null,"configMap":null,"csi":null,"downwardApi":null,"emptyDir":null,"ephemeral":null,"fc":null,"flexVolume":null,"flocker":null,"gcePersistentDisk":null,"gitRepo":null,"glusterfs":null,"hostPath":null,"iscsi":null,"nfs":null,"persistentVolumeClaim":{"$jsii.struct":{"fqn":"k8s.PersistentVolumeClaimVolumeSource","data":{"claimName":"mariadbPvc","readOnly":null}}},"photonPersistentDisk":null,"portworxVolume":null,"projected":null,"quobyte":null,"rbd":null,"scaleIo":null,"secret":null,"storageos":null,"vsphereVolume":null}}}
```

I think the ContainerProps only take volume_mounts from config files or emptyDir at this point.

misterjoshua:
Introduces PersistentVolumeClaim support. (WIP)

TODO
- [ ] PV label selectors
- [ ] PV label expressions
- [ ] Import a PVC by name
- [ ] Import a StorageClass by name
- [ ] A concrete StorageClass type (May be opening a can of worms)
- [ ] A concrete PersistentVolume type (should we?)

Signed-off-by: Josh Kellendonk <joshkellendonk@gmail.com>

Fixes #176

### DIFF
--- a/src/persistent-volume.ts
+++ b/src/persistent-volume.ts
@@ -1,0 +1,150 @@
+import { ApiObject, Size } from 'cdk8s';
+import { Construct } from 'constructs';
+import { IResource } from '.';
+import { Resource, ResourceProps } from './base';
+import * as k8s from './imports/k8s';
+
+/**
+ * Interface of a storage class.
+ */
+export interface IStorageClass extends IResource {
+}
+
+/**
+ * Access modes for persistent volumes.
+ */
+export enum PersistentVolumeAccessMode {
+  /**
+   * The volume can be mounted as read/write on one node.
+   */
+  READ_WRITE_ONCE = 'ReadWriteOnce',
+
+  /**
+   * The volume can be mounted read-only on many nodes.
+   */
+  READ_ONLY_MANY = 'ReadOnlyMany',
+
+  /**
+   * The volume can be mounted read/write on many nodes.
+   */
+  READ_WRITE_MANY = 'ReadWriteMany',
+
+  /**
+   * The volume can be mounted read/write on only one pod.
+   *
+   * This only works for CSI volumes.
+   */
+  READ_WRITE_ONCE_POD = 'ReadWriteOncePod',
+}
+
+/**
+ * Persistent volume modes.
+ */
+export enum PersistentVolumeMode {
+  /**
+   * The volume should be mounted into the pod.
+   */
+  FILESYSTEM = 'Filesystem',
+
+  /**
+   * The volume is presented to the pod as a block device.
+   */
+  BLOCK = 'Block',
+}
+
+/**
+ * Props for `PersistentVolumeClaim`
+ */
+export interface PersistentVolumeClaimProps extends ResourceProps {
+  /**
+   * Access modes for the persistent volume claim.
+   */
+  readonly accessModes: PersistentVolumeAccessMode[];
+
+  /**
+   * Storage class of the persistent volume.
+   */
+  readonly storageClass?: IStorageClass;
+
+  /**
+   * Mode of the persistent volume.
+   */
+  readonly volumeMode?: PersistentVolumeMode;
+
+  /**
+   * How to find a persistent volume claim.
+   */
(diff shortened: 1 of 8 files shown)


misterjoshua:
@Chriscbr Hey. I was fiddling with CDK8s and didn't see PersistentVolumeClaim support in cdk8s-plus-22. I've taken a crack at implementation, and I wanted to open this up for comments before I got in deep.

I've left some self review comments in the areas I suspect will need the most refinement. I'd appreciate your feedback whenever you get a chance. Thanks.

src/persistent-volume.ts
```diff
@@ -0,0 +1,150 @@
...
+import { Construct } from 'constructs';
+import { IResource } from '.';
+import { Resource, ResourceProps } from './base';
+import * as k8s from './imports/k8s';
+
+/**
+ * Interface of a storage class.
+ */
+export interface IStorageClass extends IResource {
+}
```
misterjoshua, line 11: For the sake of getting this PR open for discussion, I haven't implemented a StorageClass yet - just an interface. I went with an interface here because one might opt to use a storage class already on the k8s cluster rather than creating such a resource in their CDK8s app. This case seems common to me, so I think we should support storage class resources by name from outside of CDK8s.

src/persistent-volume.ts
```diff
@@ -0,0 +1,150 @@
...
+   * Locate a persistent volume by its name.
+   */
+  static fromVolumeName(volumeName: string): IPersistentVolumeLocation {
+    return {
+      bind: () => ({
+        volumeName,
+      }),
+    };
+  }
+
```
misterjoshua, line 109: Other possible PV locations:
* Select one by labels
* Select one by label expressions

I haven't thought what kind of syntax we should use to express PV label selection expressions. I'm open to ideas here.

misterjoshua:
Slack suggested that I should use the imports. I'm going to go with that for now and will reopen a PR with a higher-level construct if I need it again in the future. :)

PR #390, opened 2022-02-28
feat: Container resource requirements and HPA

### LINKED ISSUE #27: Container resource requirements and HPA
Containers can set their resource limits/requests.

### Use Case:
Horizontal Pod Autoscaler requirs containers to set resource requirements. 

### Proposed Solution:
```ts
const container = deployment.addContainer({
 resources: {
   cpu: {
     request: Cpu.millis(100),
     limit: Cpu.units(0.5),
   }
 }
});
container.resources.memory.request = Size.mebibytes(100);
container.resources.memory.limit = Size.gibibytes(1);
// optional? 
container.resources.setHugepage("2Mi", {
  ...
})
```

### Other:
* [x] :warning: This feature might incur a breaking change

---

mac-atre:
Allow setting compute resources (CPU and memory requests and limits) required by the container.

Signed-off-by: Mac Atre <mac.atre@gmail.com>

Fixes #27

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -1,5 +1,7 @@
+import { Size } from 'cdk8s';
 import { IConfigMap } from './config-map';
 import * as k8s from './imports/k8s';
+import type { ResourceRequirements } from './imports/k8s';
 import { Probe } from './probe';
 import { SecretValue } from './secret';
 import { Volume } from './volume';
@@ -398,6 +400,12 @@ export interface ContainerProps {
    * @default - no startup probe is defined.
    */
   readonly startup?: Probe;
+
+  /**
+   * Compute resources (CPU and memory requests and limits) required by the container
+   * @see https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+   */
+  readonly resources?: IResources;
 }
 
 /**
@@ -435,6 +443,12 @@ export class Container {
    */
   public readonly workingDir?: string;
 
+  /**
+   * Compute resources (CPU and memory requests and limits) required by the container
+   * @see https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+   */
+  public readonly resources?: IResources;
+
   private readonly _command?: readonly string[];
   private readonly _args?: readonly string[];
   private readonly _env: { [name: string]: EnvValue };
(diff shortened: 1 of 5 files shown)


src/container.ts
```diff
@@ -536,6 +551,38 @@ export class Container {
       });
     }
 
+    // Resource requests and limits
```
iliapolo, line 554: This feels a little cluttered...can you try making the flow clearer? maybe something like:

```ts
    const cpuLimit = this.resources?.cpu?.limit?.amount;
    const cpuRequest = this.resources?.cpu?.request?.amount;
    const memoryLimit = this.resources?.memory?.limit;
    const memoryRequest = this.resources?.memory?.request;

    const limits: { [key: string]: k8s.Quantity } = {};
    const requests: { [key: string]: k8s.Quantity } = {};

    if (cpuLimit) {
      limits.cpu = k8s.Quantity.fromString(cpuLimit);
    }
    if (memoryLimit) {
      limits.memory = k8s.Quantity.fromString(memoryLimit.toMebibytes().toString() + 'Mi');
    }
    if (cpuRequest) {
      requests.cpu = k8s.Quantity.fromString(cpuRequest);
    }
    if (memoryRequest) {
      requests.memory = k8s.Quantity.fromString(memoryRequest.toMebibytes().toString() + 'Mi');
    }

    const resourceRequirements = {
      limits: limits,
      requests: requests,
    };
```

?

src/container.ts
```diff
@@ -662,6 +710,47 @@ export enum MountPropagation {
   BIDIRECTIONAL = 'Bidirectional',
 }
 
+/**
+ * CPU and memory compute resources
+ */
+export interface IResources {
```
iliapolo: We normally use `IXxx` naming for [imported](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-22/main/src/secret.ts#L24) resources, not for inputs. 

Just remove the `I` prefix.

iliapolo (changes requested):

src/container.ts
mac-atre: Thanks @iliapolo. I just committed the changes above except for the following block:
```
    let resourceRequirements: ResourceRequirements | undefined = undefined;
    if (Object.keys(limits).length > 0 || Object.keys(requests).length > 0) {
      resourceRequirements = {
        limits: limits,
        requests: requests,
      };
    }
```
If I don't allow `resourceRequirements` to be undefined (when neither limits nor requests are set), it ends up being an empty object, and then the 'Instantiation properties are all respected' test fails.

iliapolo (approved):

src/container.ts
iliapolo: Fair enough 👍

PR #432, opened 2022-03-11
feat: `Role`, `ClusterRole`, `RoleBinding`, `ClusterRoleBinding` L2s

### LINKED ISSUE #24: Role L2

### LINKED ISSUE #25: ClusterRole L2
### Other:

---

### LINKED ISSUE #26: ClusterRoleBinding L2
### Other:

---

### LINKED ISSUE #374: RoleBinding L2
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding

Chriscbr:
Fixes #24 
Fixes #25
Fixes #26
Fixes #374 

Adds Role and ClusterRole L2s with "allowXxx" methods that resemble those available in the AWS CDK. This allows you to write code like:

```ts
import * as kplus from 'cdk8s-plus-23';

declare const deployment: kplus.Deployment;

const role = new kplus.ClusterRole(chart, 'my-cluster-role');
role.allowRead(deployment, kplus.ApiResource.SECRETS, kplus.ApiResource.PODS);
```

...where the "read" syntactic sugar grants permissions to the "get", "list", and "watch" commands. There are other allow commands available that offer more granularity if needed.

RoleBinding and ClusterRoleBinding L2s have also been created, and they can be instantiated through their constructors, or created using the appropriate bind methods that will attach it to the role automatically:

```ts
declare const role: kplus.Role;
declare const clusterRole: kplus.ClusterRole;

const user = new kplus.User({ name: 'alice@example.com' });
const group = new kplus.Group({ name: 'frontend-devs' });

role.bind(user, group); // creates a RoleBinding associated with the same namespace `role` was defined with
clusterRole.bindInNamespace('development', user, group); // creates a RoleBinding associated with namespace "development"
clusterRole.bind(user, group); // creates a ClusterRoleBinding, not associated with any namespace
```

Future work:
- Add full `pod.grantRead(role)` and `Resources.fromTypes(ApiResource.POD).grantRead(role)` style APIs
- Add type checking or validation to prevent adding namespaced resources to ClusterRole or cluster-wide resources to Role's etc.
- Add more constants to `ApiResource` for granting permissions to [subresources](https://stackoverflow.com/questions/57872201/how-to-refer-to-all-subresources-in-a-role-definition)
- Add convenience APIs for adding permissions to the [default roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings) using `aggregate-to-<default-role>` tags

BREAKING CHANGE: The interface `Resources` is now named `ContainerResources`.

Signed-off-by: Christopher Rybicki <rybickic@amazon.com>


iliapolo:
Suggestion for API:

```ts
const role = new Role(this, 'Role', { namespace: 'default'});
role.grantList(Grantee.fromObjects(deployment, pod));
role.grantList(Grantee.fromKind(Kind.Secrets, Kind.Deployments));
role.grantList({
  objects: [deployment, pod],
});
role.grantList({
  kinds: ['secrets', 'deployments'],
});

const binding = role.bind(user1, group1);
binding.addSubject(serviceAccount);
```

```ts
const role = new kplus.ClusterRole(chart, 'grafana-operator');
    role.addRule({
      apiGroups: [''],
      resources: ['events'],
      verbs: ['get', 'watch', 'list', 'create', 'delete', 'update', 'patch'],
    });
    role.addRule({
      apiGroups: ['integreatly.org'],
      resources: [
        'grafanadashboards',
        'grafanadatasources',
        'grafanadatasources/status',
      ],
      verbs: ['get', 'list', 'create', 'update', 'delete', 'deletecollection', 'watch'],
    });
```

Would be:

```ts
const role = new Role(this, 'Role');
role.grantGet(Kind.EVENTS);
role.grantWatch(Kind.EVENTS);
role.grantList(Kind.EVENTS);

role.grantGet(Kind.custom({ resources: 'grafanadashboards', 'grafanadatasources', apiGroup: 'integreatly.org'} 
```

src/role.ts
```diff
@@ -0,0 +1,317 @@
...
+    super(scope, id);
+
+    if (props.metadata?.namespace && props.metadata.namespace !== props.namespace) {
+      throw new Error('If `metadata.namespace` is passed as an option, its value must match the value of `namespace`.');
+    }
+
+    const metadata: ApiObjectMetadata = {
+      ...props.metadata,
+      namespace: props.namespace,
+    };
```
Chriscbr: I've exposed `namespace` as a top-level prop to make it clear that it's a required field and to make it easier to configure. But since namespace is also configurable through `metadata.namespace`, I had to add this error handling logic.

Anyway I'm open to changing this

src/grants.ts
```diff
@@ -0,0 +1,54 @@
...
+      .map(type => type.apiGroups)
+      .flat(),
+    );
+    const resources = normalizedArray(types
+      .map(type => type.resources)
+      .filter((x): x is string[] => Boolean(x))
+      .flat(),
+    );
+    return { apiGroups, resources };
+  }
```
Chriscbr: This isn't being used, since I designed `ResourceType` to implement `IGrantee` as is. Should I drop it?

src/role.ts
```diff
@@ -0,0 +1,360 @@
...
+   * @param subjects a list of subjects to bind to
+   */
+  public bind(...subjects: ISubject[]): RoleBinding {
+    const binding = new RoleBinding(this, 'RoleBinding', {
+      namespace: this.metadata.namespace!,
+      role: this,
+    });
+    binding.addSubjects(...subjects);
+    return binding;
+  }
```
Chriscbr: Note: as implemented, you can't really call any of the bind methods (`bind`, `bindInNamespace`, `bindInCluster`) more than once on the same role because you'll get an error that there are two sibling constructs with the same name.

I don't think this is a blocker (since you can just create a RoleBinding directly and give it your own name) but thought it's worth calling out.

iliapolo:
@Chriscbr I think `ApiResource` is the eureka name we've been looking for! Good idea.  

Both `pods` and a single pod named `foo` represent an `ApiResource` (i.e one that can be managed via the API) - Its just that one has a `type`, while the other has a `name`. 

So, if we define a common interface:

```ts
export interface IApiResource {
  // mandatory, both pods and a single pod has this.
  readonly apiGroup: string;

  // optional - pods will have this
  readonly type?: string;

  // optional - a single will have this
  readonly name?: string;
}
```

Then, our `Resource` class actually already implements this interface.

```ts
export abstract class Resource extends Construct implements IResource, IApiResource {
```

If we now make `ApiResource` implement it:

```ts
export class ApiResource implements IApiResource {
  /**
   * API resource information for Binding.
   */
  public static readonly BINDING = new ApiResource({
    apiGroups: [''],
    resources: ['bindings'],
  });
  ....
  ....
}
```

We can change the allow methods to accept:

```ts
public allow(verbs: string[], ...resources: IApiResource[]): void {
```

And viola, we have the API we wanted:

```ts
role.allowRead(deployment, kplus.ApiResource.SECRETS, kplus.ApiResource.PODS);
```

Thoughts?

Chriscbr:
Some more notes:
1) at the advice of Romain, I renamed `type` to `resourceType` to avoid naming conflicts in languages like Go.
2) I was re-reviewing the documentation and I think the "resources" field in rule is actually expected in all cases, as opposed to being an _alternative_ to "resourceNames". I've updated the PR based on this understanding, let me know if you have other thoughts

Ready for another round of reviews @iliapolo  :)

api-resources.txt
```diff
@@ -0,0 +1,57 @@
+NAME                              SHORTNAMES   APIVERSION                             NAMESPACED   KIND                             VERBS
```
iliapolo, line 1: Maybe add a projen task that regenerates this file given access to a cluster? It will at least document the command we need to run.

projenrc/gen-api-resource.ts
```diff
@@ -0,0 +1,284 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import { Project, SourceCode } from 'projen';
+import { snakeCase } from 'snake-case';
+
+/**
+ * Generate a source file for the `ApiResource` module, based on information about API
+ * resources generated by running `kubectl api-resources -o wide`.
```
iliapolo, line 8: Ah ok, command is here, but still, projen task could be nice :)

projenrc/gen-api-resource.ts
```diff
@@ -0,0 +1,284 @@
...
+  ts.line(' * The name of a resource type as it appears in the relevant API endpoint.');
+  ts.line(' * @example - "pods" or "pods/log"');
+  ts.line(' * @see https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources');
+  ts.line(' */');
+  ts.line('readonly resourceType: string;');
+  ts.line();
+  ts.line('/**');
+  ts.line(' * The unique, namespace-global, name of an object inside the Kubernetes cluster');
+  ts.line(' */')
+  ts.line('readonly name?: string;')
```
iliapolo: For consistency:

```suggestion
  ts.line('readonly resourceName?: string;')
```

Also maybe add something like:

*if this is omitted, the ApiResource should represent all objects of the given type.*

projenrc/gen-api-resource.ts
```diff
@@ -0,0 +1,284 @@
...
+    ts.line(` * API resource information for ${resource.kind}.`);
+    ts.line(' */');
+    ts.open(`public static readonly ${memberName} = new ApiResource({`);
+    ts.line(`apiGroup: '${apiGroups[0]}',`);
+    ts.line(`resourceType: '${resource.name}',`);
+    ts.close('});');
+    ts.line();
+  }
+
+  ts.open('public static custom(options: ApiResourceOptions): ApiResource {');
```
iliapolo, line 111: missing docstring

projenrc/gen-api-resource.ts
```diff
@@ -0,0 +1,284 @@
...
+  ts.close('}');
+  ts.line();
+  ts.line('/**');
+  ts.line(' * Represents information about an API resource type.');
+  ts.line(' */');
+  ts.open('export class ApiResource implements IApiResource {');
+
+  for (const resource of resourceTypes) {
+    const typeName = normalizeTypeName(resource.kind);
+    const memberName = snakeCase(typeName.replace(/[^a-z0-9]/gi, '_')).split('_').filter(x => x).join('_').toUpperCase();
```
iliapolo: Can we do something to transform this to a plural form?
I'd rather we have `ApiResource.PODS` and not `ApiResource.POD`.

I see that the `NAME` field in the .txt file is already plural, so maybe:

```console
memberName = <what we have now> + toUpperCase(NAME - KIND)
```

?

src/config-map.ts
```diff
@@ -52,14 +52,21 @@ export class ConfigMap extends Resource implements IConfigMap {
    * @param name The name of the config map to import
    */
   public static fromConfigMapName(name: string): IConfigMap {
-    return { name };
+    return {
```
iliapolo, line 64: You can reuse `k8s.KubeConfigMap.GVK`

src/container.ts
```diff
@@ -520,7 +520,7 @@ export interface ContainerProps {
    * Compute resources (CPU and memory requests and limits) required by the container
    * @see https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
    */
-  readonly resources?: Resources;
+  readonly resources?: ContainerResources;
```
iliapolo, line 522: 👍

src/pv.ts
```diff
@@ -92,14 +92,21 @@ export class PersistentVolume extends Resource implements IPersistentVolume, ISt
    * @param volumeName The name of the pv to reference.
    */
   public static fromPersistentVolumeName(volumeName: string): IPersistentVolume {
-    return { name: volumeName };
+    return {
```
iliapolo, line 94: `k8s.KubePersistentVolume.GVK`

src/pvc.ts
```diff
@@ -85,14 +85,21 @@ export class PersistentVolumeClaim extends Resource implements IPersistentVolume
    * @param claimName The name of the pvc to reference.
    */
   public static fromClaimName(claimName: string): IPersistentVolumeClaim {
-    return { name: claimName };
+    return {
```
iliapolo, line 87: `k8s.KubePersistentVolumeClaim.GVK`

src/secret.ts
```diff
@@ -59,14 +59,21 @@ export class Secret extends Resource implements ISecret {
    * @param name The name of the secret to reference.
    */
   public static fromSecretName(name: string): ISecret {
-    return { name };
+    return {
```
iliapolo, line 78: `k8s.KubeSecret.GVK`

src/service-account.ts
```diff
@@ -49,14 +42,21 @@ export class ServiceAccount extends Resource implements IServiceAccount {
    * @param name The name of the service account resource.
    */
   public static fromServiceAccountName(name: string): IServiceAccount {
-    return { name: name };
+    return {
```
iliapolo, line 55: `k8s.KubeServiceAccount.GVK`

src/service-account.ts
```diff
@@ -6,21 +6,14 @@ import { ISecret } from './secret';
 import { undefinedIfEmpty } from './utils';
 
 
-/**
- * Properties for initialization of `ServiceAccount`.
- */
-export interface ServiceAccountProps extends ResourceProps {
```
iliapolo: Thank you

src/service.ts
```diff
@@ -165,7 +165,7 @@ export class Service extends Resource {
   /**
    * Determines how the Service is exposed.
    */
-  public readonly type: ServiceType;
+  public readonly serviceType: ServiceType;
```
iliapolo: I think this is leftover from when `IApiResource` had a `type` property? We don't have to change this now.

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+
+}
+
+/**
+ * Options for `Role`.
+ */
+export interface RoleProps extends RoleCommonProps {
+  /**
+   * The namespace of the resources this role's permissions should apply to
+   * (required).
```
iliapolo:
```suggestion
   * The namespace this role is created in. Permissions added to this role must refer to resources in the same namespace
```

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+
+/**
+ * A reference to any Role or ClusterRole.
+ */
+export interface IRole extends IResource {
+
+}
+
+/**
+ * Options for `Role`.
```
iliapolo:
```suggestion
 * Properties for `Role`.
```

test/role.test.ts
```diff
@@ -0,0 +1,406 @@
...
+  test('with a custom rule', () => {
+
+    // GIVEN
+    const chart = Testing.chart();
+
+    // WHEN
+    const role = new kplus.Role(chart, 'pod-reader', {
+      namespace: 'default',
+    });
+    role.addRule({
```
iliapolo: You should also assert the api of the returned `PolicyRule` instance.

src/role.ts
```diff
@@ -0,0 +1,380 @@
+import { ApiObject, ApiObjectMetadata, Lazy, Names } from 'cdk8s';
+import { Construct } from 'constructs';
+import { IApiResource } from './api-resource.generated';
+import { IResource, Resource, ResourceProps } from './base';
+import * as k8s from './imports/k8s';
+import { ClusterRoleBinding, ISubject, RoleBinding } from './role-binding';
+import { filterUndefined, undefinedIfEmpty } from './utils';
+
+export interface RoleCommonProps extends ResourceProps {
+  readonly rules?: PolicyRuleProps[];
```
iliapolo: docstring

src/role.ts
```diff
@@ -0,0 +1,380 @@
+import { ApiObject, ApiObjectMetadata, Lazy, Names } from 'cdk8s';
+import { Construct } from 'constructs';
+import { IApiResource } from './api-resource.generated';
+import { IResource, Resource, ResourceProps } from './base';
+import * as k8s from './imports/k8s';
+import { ClusterRoleBinding, ISubject, RoleBinding } from './role-binding';
+import { filterUndefined, undefinedIfEmpty } from './utils';
+
+export interface RoleCommonProps extends ResourceProps {
```
iliapolo: docstring

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+  readonly rules?: PolicyRuleProps[];
+}
+
+export abstract class RoleBase extends Resource implements IResource {
```
iliapolo: docstring

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+
+  /**
+   * NonResourceURLs is a set of partial urls that a user should have access to.
+   * *s are allowed, but only as the full, final step in the path. Since
+   * non-resource URLs are not namespaced, this field is only applicable for
+   * ClusterRoles referenced from a ClusterRoleBinding. Rules can either apply
+   * to API resources (such as "pods" or "secrets") or non-resource URL paths
+   * (such as "/api"),  but not both.
+   */
+  readonly nonResourceUrLs?: string[];
```
iliapolo: The docs state:

> this field is only applicable for ClusterRoles referenced from a ClusterRoleBinding

This means this property shouldn't be available when defining a `Role`, only a `ClusterRole`. 
I would create a `RolePolicyRule` and a `ClusterRolePolicyRule` that extends it. 

Also

> Rules can either apply to API resources (such as "pods" or "secrets") or non-resource URL paths (such as "/api"),  but not both.

This means `ClusterRole` will have to validate only one of them is used. (A `Role` won't have to validate since this option won't even be available to it).

iliapolo: The docs state:

> this field is only applicable for ClusterRoles referenced from a ClusterRoleBinding

This means this property shouldn't be available when defining a `Role`, only a `ClusterRole`. 
I would create a `RolePolicyRule` and a `ClusterRolePolicyRule` that extends it. 

Also

> Rules can either apply to API resources (such as "pods" or "secrets") or non-resource URL paths (such as "/api"),  but not both.

This means `ClusterRole` will have to validate only one of them is used. (A `Role` won't have to validate since this option won't even be available to it).

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+   * will be automatically filled into this ClusterRole's rules.
+   */
+  readonly aggregationLabels?: { [key: string]: string };
+}
+
+/**
+ * ClusterRole is a cluster level, logical grouping of PolicyRules that can be
+ * referenced as a unit by a RoleBinding or ClusterRoleBinding.
+ */
+export class ClusterRole extends RoleBase implements IClusterRole {
```
iliapolo: I think this deserves its own file, since its a distinct resource.

src/role.ts
iliapolo: The fact `namespace` is required wasn't sitting right with me, so I tried to create a role without a namespace and apply - it worked...

I'm not sure why the docs explicitly say:

> A Role always sets permissions within a particular [namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces); when you create a Role, you have to specify the namespace it belongs in.

But I would actually skip this for now, its awkward and i'm not sure its worth it. Let's just stick to the normal namespace handling we have for all resources.

test/role.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      {
+        apiGroups: [''],
+        resources: ['pods/log'],
+        resourceNames: [],
+        verbs: ['get', 'list', 'watch'],
+      },
+    ]));
+
+  });
+
```
iliapolo, line 226: I think a mixed test is appropriate as well, e.g:

```ts
role.allowRead(deployment, kplus.ApiResource.POD);
```

test/role-binding.test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  const role = new kplus.Role(chart, 'pod-reader', {
+    namespace: 'default',
+  });
+  role.allowRead(kplus.ApiResource.POD);
+
+  // WHEN
+  const user = new kplus.User({ name: 'alice@example.com' });
+  const group = new kplus.Group({ name: 'frontend-admins' });
+  role.bind(user, group);
```
iliapolo: Add assertions on the `RoleBinding` api that is returned.

src/role.ts
iliapolo: I think this is ok for `bind` and `bindInCluster`, but not so much for `bindInNamespace`, see relevant comment.

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+  }
+
+  /**
+   * Create a RoleBinding that binds the permissions in this ClusterRole
+   * to a list of subjects, that will only apply to the given namespace.
+   * @param namespace the namespace to limit permissions to.
+   * @param subjects a list of subjects to bind to
+   */
+  public bindInNamespace(namespace: string, ...subjects: ISubject[]): RoleBinding {
+    const binding = new RoleBinding(this, 'RoleBinding', {
```
iliapolo: This is problematic. From the docs:

> This kind of reference lets you define a set of common roles across your cluster, then reuse them within multiple namespaces.

i.e, calling this method multiple times is actually a common use case. Fortunately there's an easy fix:

```suggestion
    const binding = new RoleBinding(this, `RoleBinding${namespace}`, {
```

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+    binding.addSubjects(...subjects);
+    return binding;
+  }
+
+  /**
+   * Create a ClusterRoleBinding that binds the permissions in this
+   * ClusterRole to a list of subjects, without namespace restrictions.
+   * @param subjects a list of subjects to bind to
+   */
+  public bindInCluster(...subjects: ISubject[]): ClusterRoleBinding {
```
iliapolo: I think we can use `bind` here. 

```suggestion
  public bind(...subjects: ISubject[]): ClusterRoleBinding {
```

This way its consistent with the `bind` API in `Role`. So we have:

- `role.bind -> RoleBinding`
- `clusterRole.bind -> ClusterRoleBinding`
- `clusterRole.bindInNamespace` -> `RoleBinding`.

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+   * The role to bind to.
+   */
+  readonly role: IClusterRole;
+}
+
+/**
+ * A ClusterRoleBinding grants permissions cluster-wide to a user or
+ * set of users.
+ */
+export class ClusterRoleBinding extends Resource {
```
iliapolo, line 127: I think this deserves its own file, since its a distinct resource.

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+      namespace: props.namespace,
+    };
+
+    this.apiObject = new k8s.KubeRole(this, 'Resource', {
+      metadata,
+      rules: Lazy.any({ produce: () => this.synthesizeRules() }),
+    });
+  }
+
+  private synthesizeRules(): any {
```
iliapolo:
```suggestion
  private synthesizeRules(): k8s.PolicyRule[] {
```

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+   */
+  readonly verbs: string[];
+}
+
+/**
+ * Information that describes a policy rule, but does not contain information
+ * about who the rule applies to or which namespace the rule applies to.
+ */
+export class PolicyRule {
+  constructor(public readonly props: PolicyRuleProps) {}
```
iliapolo: We normally don't just expose `props` as a public property. Flatten these out to the individual properties it contains - makes for a simpler API for consumers.

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+  /**
+   * Namespace of the referenced object.  If the object kind is non-namespace,
+   * such as "User" or "Group", and this value is not empty the Authorizer
+   * should report an error.
+   */
+  readonly namespace?: string;
+}
+
+/**
+ * Options for `RoleBinding`.
```
iliapolo:
```suggestion
 * Properties for `RoleBinding`.
```

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+export interface RoleBindingProps extends ResourceProps {
+  /**
+   * The role to bind to. A RoleBinding can reference a Role or a ClusterRole.
+   */
+  readonly role: IRole;
+
+  /**
+   * Namespace of the resources the RoleBinding should apply to.
+   */
+  readonly namespace: string;
```
iliapolo: I dont think this should be required. The default should be to use the role's namespace (even if its undefined). 

I also wonder what happens if the `namespace` is different than that of the role's, is that allowed? can you verify by trying to apply such a manifest?

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+      name: subject.name,
+      namespace: subject.namespace,
+    }));
+  }
+}
+
+/**
+ * Options for `User`.
+ */
+export interface GroupProps {
```
iliapolo:
```suggestion
export interface UserProps {
```

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+      apiGroup: subject.apiGroup === 'core' ? '' : subject.apiGroup,
+      kind: subject.kind,
+      name: subject.name,
+      namespace: subject.namespace,
+    }));
+  }
+}
+
+/**
+ * Options for `User`.
```
iliapolo:
```suggestion
 * Properties for `User`.
```

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+ */
+export interface GroupProps {
+  /**
+   * The name of the user.
+   */
+  readonly name: string;
+  /**
+   * The namespace that the group applies to.
+   */
+  readonly namespace?: string;
```
iliapolo: Something doesn't add up, the schema docs for this property say:

> If the object kind is non-namespace, such as "User" or "Group", and this value is not empty the Authorizer should report an error.

Are you sure a group is a namespaced resource? I think this property is only relevant for a `ServiceAccount` subject.

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+    this.name = props.name;
+  }
+}
+
+/**
+ * Options for `Group`.
+ */
+export interface GroupProps {
+  /**
+   * The name of the user.
```
iliapolo:
```suggestion
   * The name of the group.
```

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+  public readonly apiGroup: string | undefined = 'rbac.authorization.k8s.io';
+  public readonly kind: string = 'User';
+  public readonly name: string;
+  constructor(props: GroupProps) {
+    this.name = props.name;
+  }
+}
+
+/**
+ * Options for `Group`.
```
iliapolo:
```suggestion
 * Properties for `Group`.
```

src/role-binding.ts
```diff
@@ -0,0 +1,247 @@
...
+import * as k8s from './imports/k8s';
+import { IClusterRole, IRole } from './role';
+import { filterUndefined } from './utils';
+
+/**
+ * Subject contains a reference to the object or user identities a role binding
+ * applies to.  This can either hold a direct API object reference, or a value
+ * for non-objects such as user and group names.
+ */
+export interface ISubject {
```
iliapolo, line 13: I expected to see `ServiceAccount` implement this interface. Is this something you didn't get to?

src/role.ts
```diff
@@ -0,0 +1,380 @@
...
+      rules: Lazy.any({ produce: () => this.synthesizeRules() }),
+      aggregationRule: Lazy.any({ produce: () => this.synthesizeAggregationRules() }),
+    });
+  }
+
+  /**
+   * Add the rules of the argument ClusterRole into this ClusterRole.
+   * @param role
+   */
+  public aggregateFrom(role: ClusterRole): void {
```
iliapolo: I'm not sure about this API. Does an aggregation only makes sense with another managed role? What if I just need to add an aggregation label, pointing to an existing `ClusterRole`?

How about:

```ts
// aggregate rules from roles matching this label selector
public aggregate(key: string, value: string): void;

// combine rules from the given role by applying the appropriate aggregation labels
public combine(role: ClusterRole): void;
```

iliapolo (changes requested):
Sweet! This is a big one.

projenrc/gen-api-resource.ts
Chriscbr: Good idea. I had to do something slightly more involved because some English plurals are a bit funky (like "capacity" -> "capacities") but it does work 👍

Chriscbr:
Still WIP

src/role.ts
Chriscbr: Interesting, this is a good catch.

I was thinking about this for a bit, and I realized if we say a `RolePolicyRule` is a rule about API resources, and a `ClusterRolePolicyRule` is a rule about API resources OR non-resources, it actually makes more sense for the subtyping relation to go the other way around -- that is, `RolePolicyRule` is a subtype (a special case of) `ClusterRolePolicyRule`!

To make the subtyping relationship clearer I've renamed them `PolicyRule` and `ResourcePolicyRule` respectively (see updated PR). Thoughts?

src/role.ts
Chriscbr: I'd rather leave this as is since to get some inheritance behavior working I need `Role` and `ClusterRole` to both have access to the un-exported `RoleBase` class (otherwise JSII throws errors preventing me from giving `Role.addRule(rule: ResourcePolicyRuleProps)` a different signature from `RoleBase.addRule(rule: PolicyRuleProps)`).

src/role-binding.ts
Chriscbr: To align with `roles.ts` I'll leave this as is

src/role-binding.ts
Chriscbr: Oops

src/role-binding.ts
Chriscbr: Yep I believe you're correct

src/role-binding.ts
Chriscbr: Ah I had a test for this, but it was only working because of structural typing. Fixed

src/role.ts
Chriscbr: Good idea 👍

Chriscbr:
Thanks for all of the feedback @iliapolo ! This should be ready for another review 🙂

iliapolo (approved):

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-21/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/685)|

### Questions ?

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-20/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/686)|

### Questions ?

PR #455, opened 2022-03-16
feat: security context

### LINKED ISSUE #348: Configure the security context of containers and pods
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

iliapolo:
`SecurityContext` for pods and containers. I included only the properties that I thought were most common and that I could quickly understand. If we see the need to add additional ones, we'll do that at a later time. 

The default values correspond to the values applied by Kubernetes if the property is not specified.

Fixes https://github.com/cdk8s-team/cdk8s-plus/issues/348


iliapolo:
@Chriscbr Even though `PodSecurityContext` and `ContainerSecurityContext` share some common properties, I intentionally didn't create a common interface/class since it felt the coupling is not worth it compared to the small duplication.

src/container.ts
```diff
@@ -6,6 +6,81 @@ import { Probe } from './probe';
...
+
+  constructor(props: ContainerSecurityContextProps = {}) {
+    this.ensureNonRoot = props.ensureNonRoot ?? false;
+    this.privileged = props.privileged ?? false;
+    this.readOnlyRootFilesystem = props.readOnlyRootFilesystem ?? false;
+    this.user = props.user;
+    this.group = props.group;
+  }
+
+  public _toKube(): k8s.SecurityContext {
```
Chriscbr, line 75: Mark as @ internal

src/pod.ts
```diff
@@ -387,6 +469,46 @@ export class Pod extends Resource implements IPodSpec {
...
+  private readonly _sysctls: Sysctl[] = [];
+
+  constructor(props: PodSecurityContextProps = {}) {
+    this.ensureNonRoot = props.ensureNonRoot ?? false;
+    this.fsGroupChangePolicy = props.fsGroupChangePolicy ?? FsGroupChangePolicy.ALWAYS;
+    this.user = props.user;
+    this.group = props.group;
+    this.fsGroup = props.fsGroup;
+
+    (props.sysctls ?? []).forEach(s => this._sysctls.push(s));
```
Chriscbr: style: prefer for loops over forEach

src/pod.ts
```diff
@@ -209,6 +211,7 @@ export class PodSpec implements IPodSpec {
       restartPolicy: this.restartPolicy,
       serviceAccountName: this.serviceAccount?.name,
       containers: containers,
+      securityContext: this.securityContext._toKube(),
```
Chriscbr, line 214: Hmm. Does this always get rendered on all pods and podSpec's now? If so I feel like this should have been caught by a snapshot test of some kind.

test/pod.test.ts
```diff
@@ -290,3 +290,55 @@ test('automatically adds volumes from init container mounts', () => {
   expect(spec.volumes[0].emptyDir).toBeTruthy();
 
 });
+
+test('default security context', () => {
```
Chriscbr, line 294: nit: Maybe we can update this file with GIVEN / WHEN / THEN comments - I think it's a nice pattern to help structure tests for contributors 🙂

src/pod.ts
```diff
@@ -387,6 +469,46 @@ export class Pod extends Resource implements IPodSpec {
...
+/**
+ * Holds pod-level security attributes and common container settings.
+ */
+export class PodSecurityContext {
+
+  public readonly ensureNonRoot: boolean;
+  public readonly user?: number;
+  public readonly group?: number;
+  public readonly fsGroup?: number;
+  public readonly fsGroupChangePolicy: FsGroupChangePolicy;
```
Chriscbr, line 484: Are `ensureNonRoot` and `fsGroupChangePolicy` required here because they are required in the kubernetes manifest? Or because we are suggesting them as opinionated defaults for all pods / pod templates?

I imagine an alternative could be making all security fields optional, so that if no PodSecurityContext is provided to `PodSpecProps`, then the pod `securityContext` doesn't get rendered.

If we want to keep these defaults, perhaps you can add a note in the PR description to just explain why these defaults were chosen (since I think it's fair to say "sane defaults" is one the main features of cdk8s+. I imagine we have design tenets about this somewhere 😛).

src/pod.ts
```diff
@@ -254,6 +257,74 @@ export class PodTemplate extends PodSpec implements IPodTemplate {
...
+  readonly fsGroup?: number;
+
+  /**
+    * Defines behavior of changing ownership and permission of the volume before being exposed inside Pod.
+    * This field will only apply to volume types which support fsGroup based ownership(and permissions).
+    * It will have no effect on ephemeral volume types such as: secret, configmaps and emptydir.
+    *
+    * @default FsGroupChangePolicy.ALWAYS
+    */
+  readonly fsGroupChangePolicy?: FsGroupChangePolicy;
```
Chriscbr: nit: Docstrings *'s should be indented the same

Chriscbr (approved):
This looks great @iliapolo ! 😄 I'll approve this with `do-not-merge` so you can look at the comments and address them as you see fit.

src/pod.ts
iliapolo: Yeah snapshot didn't get updated cause of other build failures 👍

test/pod.test.ts
iliapolo: Meh, I also used to like it but find it a bit ceremonial now

src/pod.ts
iliapolo: > Or because we are suggesting them as opinionated defaults for all pods / pod templates?

Yes defaults, but not opinionated since they simply reflect the defaults Kubernetes applies.
It makes it easier for callers since they don't have to worry about the undefined case when inspecting these attributes.

> If we want to keep these defaults, perhaps you can add a note in the PR description to just explain why these defaults were chosen

👍

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-21/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/561)|

### Questions ?

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-20/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/566)|

### Questions ?

PR #497, opened 2022-03-30
feat(pod): pod dns settings

### LINKED ISSUE #362: Customize pod DNS settings
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/

iliapolo:
Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/362


src/pod.ts
```diff
@@ -50,6 +50,16 @@ export interface IPodSpec {
...
 
+  /**
+   * The pod's DNS settings.
+   */
+  readonly dns: PodDns;
+
+  /**
+   * The pod's security context.
+   */
+  readonly securityContext: PodSecurityContext;
```
iliapolo, line 62: I just noticed this was missing for the security context PR.

src/pod.ts
```diff
@@ -234,6 +248,11 @@ export class PodSpec implements IPodSpec {
       initContainers: initContainers,
       hostAliases: this.hostAliases,
       volumes: Array.from(volumes.values()).map(v => v._toKube()),
+      dnsPolicy: dns.policy,
+      dnsConfig: dns.config,
+      hostname: dns.hostname,
+      subdomain: dns.subdomain,
+      setHostnameAsFqdn: dns.hostnameAsFQDN,
```
Chriscbr, line 258: I dig this abstraction

src/pod.ts
```diff
@@ -507,6 +541,186 @@ export class Pod extends Resource implements IPodSpec {
...
+  public addSearch(...searches: string[]) {
+    this._searches.push(...searches);
+  }
+
+  /**
+   * Add a custom option.
+   */
+  public addOption(...options: DnsOption[]) {
+    this._options.push(...options);
+  }
```
Chriscbr, line 701: Would it make more sense to store this as a dictionary? and perhaps make this method could accept `key: string, value: string`? Also if we're using an array, should we be de-duplicating options with the same keys?

Chriscbr (approved):
LGTM, I just left one small question - flagged this with `do-not-merge` so you can take a look and update it or leave it as is. 👍

src/pod.ts
iliapolo: The problem is that the `value` of an option is optional (basically flags). This makes a key-value API awkward when provided in the properties, since it would require something like:

`{ op1: undefined }`

Meh :\, I'll leave it as is. 

As for deduping, maybe...it might incur some unexpected behavior though. i'll leave it for later consideration.

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-21/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/636)|

### Questions ?

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-20/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/637)|

### Questions ?

PR #655, opened 2022-04-27
chore: fix backport workflow

iliapolo:
There's currently somewhat of a mess with the backport workflow, which doesn't actually work. 

The action we use is [tibdex/backport@v1](https://github.com/tibdex/backport), which I mistakenly thought was internally using the [backport](https://github.com/sqren/backport) tool i've been testing this out with. The correct action to use is actually [sqren/backport-github-action](https://github.com/sqren/backport-github-action). 

However, the action doesn't currently use the correct version of `backport` that contains the [fix](https://github.com/sqren/backport/pull/336) we need for DCO to pass.

I asked the author to cut a new release, but he hasn't responded yet. I then realized the action is not really necessary, since most of the logic is implemented in the `backport` tool (finally a well moduled action). So this PR configures the workflow to use the `backport` tool as a normal `run` step, via projen.  

Signed-off-by: Eli Polonsky <epolon@amazon.com>


.projenrc.js
```diff
@@ -115,29 +115,42 @@ for (const lang of ['typescript', 'python', 'java']) {
   project.gitignore.exclude(output);
 }
 
+// backport task to branches based on pr labels
+const backportTask = project.addTask('backport');
+backportTask.exec('npx backport --accesstoken ${GITHUB_TOKEN} --pr ${BACKPORT_PR_NUMBER} --non-interactive');
```
Chriscbr: It looks like these environment variables might be required. If so we can add a runtime check for this:
```suggestion
// backport task to branches based on pr labels
const backportTask = project.addTask('backport', {
  requiredEnv: ['GITHUB_TOKEN', 'BACKPORT_PR_NUMBER'],
});
backportTask.exec('npx backport --accesstoken ${GITHUB_TOKEN} --pr ${BACKPORT_PR_NUMBER} --non-interactive');
```

.projenrc.js
```diff
@@ -147,6 +160,15 @@ const backportConfig = {
...
   signoff: true,
+  branchLabelMapping: {
+    '^backport-to-(.+)$': '$1',
+  },
+  prTitle: '{commitMessages}',
+  fork: false,
+  publishStatusCommentOnFailure: true,
+  publishStatusCommentOnSuccess: true,
+  publishStatusCommentOnAbort: true,
+  targetPRLabels: [project.autoApprove.label],
```
Chriscbr, line 172: Where are these options documented? (add comment/link)

Chriscbr (approved):
Should we also update the contributing guide?

.projenrc.js
iliapolo: Ohh nice, didn't know about that - thanks!

.projenrc.js
iliapolo: Good call out - i'll add

PR #659, opened 2022-04-27
feat(stateful-set): `minReady` support

### LINKED ISSUE #373: Customize StatefulSet minReadySeconds
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates

iliapolo:
Fixes https://github.com/cdk8s-team/cdk8s-plus/issues/373


src/stateful-set.ts
```diff
@@ -49,6 +49,18 @@ export interface StatefulSetProps extends workload.WorkloadProps {
    * @default - RollingUpdate with partition set to 0
    */
   readonly strategy?: StatefulSetUpdateStrategy;
+
+  /**
+   * Minimum duration for which a newly created pod should be ready without any of its container crashing,
+   * for it to be considered available. Zero means the pod will be considered available as soon as it is ready.
+   *
+   * This is an alpha field and requires enabling StatefulSetMinReadySeconds feature gate.
```
Chriscbr, line 57: Will it be a problem if the feature gate is not enabled and we are still including `minReadySeconds` in the generated template? Might require manual testing to validate this works as expected, not sure

Chriscbr (approved):

PR #660, opened 2022-04-27
feat(workload): expression selectors and pod metadata integration

### LINKED ISSUE #413: Add deployment selector as pod metadata
Currently, adding a selector to a deployment takes place like so:

```ts
deployment.selectByLabel('app', 'app');
```

Which will create:

```yaml
 selector:
    matchLabels:
      app: app
```

A common scenario of adding selectors to a deployment also involves adding the same selector as a pod metadata label. 
Actually, we already assume this scenario and provide a built-in default selector that does this:

https://github.com/cdk8s-team/cdk8s-plus/blob/d9e33424e13f85fd9e1b18494a0a013bc6905d8f/src/deployment.ts#L138-L143

The proposal is to extend the `selectByLabel` method so it can also add the label to the pod metadata.

### LINKED ISSUE #414: Expose default selector configuration on `Deployment`
Currently the `Deployment` construct creates a default selector and applies it to the pod metadata.

The problem is that sometimes custom selector configurations need to be passed. It would be nice to have a `defaultSelectors` option that accept labels and applies the same logic.

iliapolo:
Introduce a new API for workload selectors that supports both `matchExpressions` as well as applying the entries in `matchLabels` to the pod metadata for automatic selection.

BREAKING CHANGE: `defaultSelector` renamed to `select`. `selectByLabel` renamed to `select` and underwent an API change: `deployment.selectByLabel('key', 'value') -> deployment.select(LabelSelector.equals('key', 'value'))`

Fixes https://github.com/cdk8s-team/cdk8s-plus/issues/414
Fixes https://github.com/cdk8s-team/cdk8s-plus/issues/413


src/workload.ts
```diff
@@ -13,14 +13,65 @@ export interface WorkloadProps extends AbstractPodProps {
...
    * @default true
    */
-  readonly defaultSelector?: boolean;
+  readonly select?: boolean;
 
 }
 
+/**
+ * Possible operators.
+ */
```
Chriscbr, line 28: Link to k8s docs?

Chriscbr (approved):

PR #678, opened 2022-04-30
chore: roles refactor

iliapolo:
@Chriscbr Following our conversation, I think we need to take a somewhat different approach. 

We've been trying to optimize for code maintenance by creating a common `RoleBase` that gives us some common functionality between a `Role` and a `ClusterRole`. However, we should be optimizing for user experience. 

A `Role` can only define rules for resources, where as a `ClusterRole` can define rules both for resources and non-resources. Trying to express this inherent difference, while still keeping a common base class, is I think what gets us in trouble. The solution we came up with is to have a common `addResourceRule` method in the base class, and have `ClusterRole` add an `addNonResourceRule` method that will only apply to it. 

This will result in the following API:

```ts
const role = new kplus.Role(...);
role.addResourceRule(...);
role.allow(...);

const clusterRole = new kplus.Role(...);
clusterRole.addResourceRule(...);
clusterRole.addNonResourceRule(...);
clusterRole.allow(...);
```

There are several problems with this:

- Naming inconsistency (`addResourceRule` vs `allow` - even though both refer to resource rules). Note that if we change `addResourceRule -> addRule`, it won't be clear why `ClusterRole` has `addRule` and `addNonResourceRule`.
- The `add` methods don't offer any capabilities that the `allow` method don't. It creates a situation where we have two API's offering the same thing - which is confusing. I'm not sure the `add` methods have merit in higher level API's, the whole point is to remove this bulky k8s structure. If users want to use the underlying structure instead, they can always escape hatch. 

If we focus on optimizing user experience for the common use-case, I think the API should look like this: 

```ts
const role = new kplus.Role(...);
// allow reading a specific pod and all secrets
role.allowGet(pod, ApiResource.SECRETS);

const clusterRole = new kplus.ClusterRole(...);
// allow reading a specific pod, all secrets, and the /healthz endpoint
clusterRole.allowGet(pod, ApiResource.SECRETS, ApiEndpoint.of('/healthz'));
```

Note that a `Role` can be given only api resources, but a `ClusterRole` can be given non api resources as well, all with the same API. 

This PR adds an `ApiEndpoint` (this name is also used in k8s docs: *non-resource endpoints (like /healthz)*) interface that allows `ClusterRole` to accept everything it can. It also removes the `add` methods in favor of the existing `allow` ones. 

Note that the base was class was completely removed, which of course resulted in a some duplication, but I think its worth it for the flexibility it gives. 

#### Resources

- https://octopus.com/blog/k8s-rbac-roles-and-bindings
- https://kubernetes.io/docs/reference/access-authn-authz/rbac/


src/role.ts
```diff
@@ -1,34 +1,100 @@
...
-    this._rules = [];
+    this.apiObject = new k8s.KubeRole(this, 'Resource', {
+      metadata: props.metadata,
+      rules: Lazy.any({ produce: () => this.synthesizeRules() }),
+    });
   }
 
-  public abstract addRule(rule: PolicyRuleProps): PolicyRule;
+  /**
+   * Rules associaated with this Role.
```
Chriscbr, line 81:
```suggestion
   * Rules associated with this Role.
```

src/role.ts
```diff
@@ -213,282 +193,299 @@ export class Role extends RoleBase implements IRole {
...
+
+  private synthesizeRules(): k8s.PolicyRule[] {
+    const rules: k8s.PolicyRule[] = [];
+    for (const rule of this._rules) {
+      for (const resource of rule.resources) {
+        rules.push({
+          verbs: rule.verbs,
+          apiGroups: [resource.apiGroup === 'core' ? '' : resource.apiGroup],
+          resourceNames: resource.resourceName ? [resource.resourceName] : undefined,
+          resources: resource.resourceType ? [resource.resourceType] : undefined,
```
Chriscbr, line 205: nit: I think we might be able to simplify this
```suggestion
          resources: resource.resourceType,
```

test/role.test.ts
```diff
@@ -312,42 +319,20 @@ Array [
 
     // WHEN
     const role = new kplus.ClusterRole(chart, 'pod-reader');
-    const rule = role.addRule({
-      nonResourceUrls: ['/healthz', '/healthz/*'],
-      verbs: ['get', 'post'],
-    });
+    role.allow(['get', 'post'], kplus.NonApiResource.of('/healthz'), kplus.NonApiResource.of('/healthz/*'));
```
Chriscbr, line 322: Maybe we could allow adding multiple endpoints through a single call? It's a possible future enhancement anyways 🙂 
```suggestion
    role.allow(['get', 'post'], kplus.NonApiResource.of('/healthz', '/healthz/*'));
```

Chriscbr (approved):
Great improvements, I think the API makes a lot more sense now. It's true, there's some perceived duplication between the classes, but we are able to give the users more granular typing between the arguments of them (IApiResource vs IApiEndpoint) and I think these interface names are appropriately descriptive as well. 👍

PR #684, opened 2022-05-04
feat(pod, workload): scheduling strategy (`affinity`, `tolerations`, `nodeName`)

### LINKED ISSUE #31: Add support for affinity rules in pods and deployments
Please allow support to add affinity for deployment template specification.

### LINKED ISSUE #360: Constrain the nodes that pods run on through node selectors and nodeName
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector

### LINKED ISSUE #377: Constrain the nodes that pods run on through taints and tolerations
https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

### LINKED ISSUE #689: `Namespace` L2
https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/

iliapolo:
Docs are here: https://github.com/cdk8s-team/cdk8s/pull/911

### A few notes:

- The API doesn't use the terminology used by the k8s properties. Instead, it uses terminology from the Kubernetes [docs](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/), in an attempt to more clearly convey the intent of these capabilities. Would love input on this choice.
- The API isn't exposed as constructor properties. I felt the semantic post-instantiation methods are way better at conveying the intent in this case. We can definitely add it though is there will be demand. 

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/31
Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/377
Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/360
Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/689


src/pod.ts
```diff
@@ -720,3 +785,535 @@ export interface HostAlias {
...
+
+  private constructor(
+    public readonly key: string,
+    public readonly operator: string,
+    public readonly values?: string[]) {
+  }
+}
+
+/**
+ * Label queries that can be perofmed against pods.
```
Chriscbr:
```suggestion
 * Represents a query that can be performed against pod labels.
```

test/pod.test.ts
```diff
@@ -521,3 +521,239 @@ test('auto mounting token can be disabled', () => {
...
+
+  });
+
+  test('can tolerate tainted nodes', () => {
+
+    const chart = Testing.chart();
+
+    const devNodes = kplus.Node.tainted(
+      kplus.NodeTaintQuery.is('key1', 'value1'),
+      kplus.NodeTaintQuery.is('key2', 'value2', { effect: kplus.TainEffect.PREFER_NO_SCHEDULE }),
```
Chriscbr:
```suggestion
      kplus.NodeTaintQuery.is('key2', 'value2', { effect: kplus.TaintEffect.PREFER_NO_SCHEDULE }),
```

test/pod.test.ts
```diff
@@ -521,3 +521,239 @@ test('auto mounting token can be disabled', () => {
...
+  });
+
+  test('can be assigned to a node by name', () => {
+
+    const chart = Testing.chart();
+
+    const redis = new kplus.Pod(chart, 'Redis', { containers: [{ image: 'redis' }] });
+    redis.scheduling.assign(kplus.Node.named('node1'));
+
+    expect(Testing.synth(chart)).toMatchSnapshot();
```
Chriscbr, line 573: I don't see `node1` referenced anywhere in the snapshot - is that intended?

test/pod.test.ts
```diff
@@ -521,3 +521,239 @@ test('auto mounting token can be disabled', () => {
...
+    });
+    const web = new kplus.Pod(chart, 'Web', {
+      containers: [{ image: 'web' }],
+    });
+
+    redis.metadata.addLabel('app', 'store');
+
+    web.scheduling.colocate(redis);
+
+    expect(Testing.synth(chart)).toMatchSnapshot();
```
Chriscbr: What happens if we don't explicitly add a label to `redis`? Does it throw an error, or automatically add labels to make it work, or something else?

src/workload.ts
```diff
@@ -135,55 +110,46 @@ export abstract class Workload extends AbstractPod {
...
+export class WorkloadScheduling extends pod.PodScheduling {
 
   /**
-   * Creates a `matchExpressions` "DoesNotExist" entry.
+   * Spread the pods in this workload by the topology key.
    */
-  public static doesNotExist(key: string) {
-    return new LabelSelector(false, key, undefined, LabelSelectorRequirementOperator.DOES_NOT_EXIST);
+  public spread(topologyKey: pod.Topology, options: WorkloadSchedulingSpreadOptions = {}) {
+    // a spread is a separation of the pod from itself
```
Chriscbr: optional: Move this comment into the docstring

src/workload.ts
```diff
@@ -135,55 +110,46 @@ export abstract class Workload extends AbstractPod {
...
+   * @default - no weight. spread is assumed to be required.
    */
-  public static exists(key: string) {
-    return new LabelSelector(false, key, undefined, LabelSelectorRequirementOperator.EXISTS);
-  }
+  readonly weight?: number;
+
+}
+
+export class WorkloadScheduling extends pod.PodScheduling {
```
Chriscbr, line 169: nit: Docstring

src/pod.ts
```diff
@@ -65,6 +68,13 @@ export abstract class AbstractPod extends base.Resource {
     return [...this._hostAliases];
   }
 
+  public get podSelector(): PodSelector {
+    return {
+      labelSelector: Object.keys(this.podMetadata.toJson().labels).map(l => PodLabelQuery.is(l, this.podMetadata.getLabel(l)!)),
+      namespaceSelector: this.metadata.namespace ? [PodLabelQuery.is('kubernetes.io/metadata.name', this.metadata.namespace)] : undefined,
+    };
```
Chriscbr: Add unit test for this behavior (I was expecting to find `kubernetes.io/metadata.name` in one of the snapshots)

src/pod.ts
```diff
@@ -722,3 +782,581 @@ export interface HostAlias {
...
+   */
+  public static custom(key: string): Topology {
+    return new Topology(key);
+  }
+
+  private constructor(public readonly key: string) {};
+}
+
+/**
+ * Options for `PodScheduling.coloate`.
```
Chriscbr:
```suggestion
 * Options for `PodScheduling.colocate`.
```

src/pod.ts
```diff
@@ -722,3 +782,581 @@ export interface HostAlias {
...
+    }
+  }
+
+  /**
+   * @internal
+   */
+  public _toKube(): { affinity?: k8s.Affinity; nodeName?: string; tolerations?: k8s.Toleration[] } {
+
+    const atLeastOne = (...arrays: Array<any>[]) => {
+      return [...arrays].map(a => a.length).reduce((a, b) => a + b) > 0;
```
Chriscbr: (optional) Another way to shave the yak 😜 
```suggestion
      return arrays.flat().length > 0;
```

Chriscbr (changes requested):

test/pod.test.ts
iliapolo: 👍

src/pod.ts
iliapolo: Oh nice, much obliged!

src/pod.ts
iliapolo: Outdated, should be covered now.

src/workload.ts
iliapolo: Added

src/workload.ts
iliapolo: Moved 👍

test/pod.test.ts
iliapolo: Yeah this was one of the things I wanted to change. We now automatically add `cdk8s.io/metadata.addr` label to every pod and workload resource, and use this label as a way to uniquely identify and select managed pods.

test/pod.test.ts
iliapolo: Wow - good catch. Bug fixed 👍

test/namespace.test.ts
```diff
@@ -0,0 +1,26 @@
+import { Testing, ApiObject } from 'cdk8s';
+import { Node } from 'constructs';
+import * as kplus from '../src';
+
+test('defaultChild', () => {
```
Chriscbr, line 5: Can we also add a test checking what a `new kplus.Namespace()` synthesizes to in a template? I imagine it should look something like this: https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#creating-a-new-namespace

src/namespace.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * @see INamespaceSelector.toNamespaceName()
+   */
+  public toNamespaceName(): string | undefined {
+    return this.name;
+  }
+
+  /**
+   * @see IPeer.toNamespacedPodSelector()
```
Chriscbr: Looks like IPeer was an interface that got scrapped?

src/pod.ts
```diff
@@ -362,19 +420,116 @@ export interface AbstractPodProps extends base.ResourceProps {
...
+  toPodSelector(): IPodSelector;
+
+  /**
+   * Return the label selector that selects the namespaces.
+   *
+   * - Implementors should return `undefined` to select "this" namespace (i.e the namespace of the selected pod).
+   * - Implementors should return `Namespace.all()` to select all namespaces.
+   */
+  toNamespaceSelector(): namespace.INamespaceSelector | undefined;
+}
```
Chriscbr: Could we simplify this to something like `INamespacedPodSelector extends IPodSelector, INamespaceSelector`?

Also -- from my reading of this API it looks like the user can select pods by their labels or which namespace they're in, but not necessarily by the node they're in. Can you check if my intuition on that is right? There's a lot of "selector"-ish concepts being introduced in this PR so I might be totally wrong.

iliapolo: As discussed internally, having `INamespacedPodSelector extend IPodSelector` will mean that `Pod.labeled(...).namespaced(...)` can be passed to resources expecting only `IPodSelector`.

That is, some resources don't allow namespace selection in their selector definition, and those resources simply accept an `IPodSelector`. It shouldn't be possible to pass `INamespacedPodSelector` to those resources.

src/namespace.ts
iliapolo: It actually belongs in a different PR - good catch.

test/namespace.test.ts
iliapolo: Yeap, added 👍

src/pod.ts
iliapolo: > Also -- from my reading of this API it looks like the user can select pods by their labels or which namespace they're in, but not necessarily by the node they're in. Can you check if my intuition on that is right? There's a lot of "selector"-ish concepts being introduced in this PR so I might be totally wrong.

Users can select pods by one of three methods:

1. Strict pod label matching (not modeled in this PR)
2. Pod label queries (`IPodSelector`)
3. Pod label queries and namespaces (`INamespacedPodSelector`)

There is no API saying "select a pod that runs on this node". There's only "select a node with these labels, and run the pod on it".

PR #724, opened 2022-05-20
fix(deployment): unable to set `maxUnavailable` on a rolling update strategy

nielsslot:
This looks like a copy-paste error. I couldn't create a deployment strategy with different maxSurge and maxUnavailable values because of it. But with this fix different values for maxSurge and maxUnavailable simply work.


iliapolo (changes requested):
Oh my, thanks for this! 

Can you also add a unit test in [`deployment.test.ts`](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-22/main/test/deployment.test.ts) ?

iliapolo (approved):

PR #737, opened 2022-05-22
feat: network policies

### LINKED ISSUE #446: `NetworkPolicy` L2
https://kubernetes.io/docs/concepts/services-networking/network-policies/

iliapolo:
Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/446

Docs: https://github.com/cdk8s-team/cdk8s/pull/925


src/network-policy.ts
```diff
@@ -0,0 +1,517 @@
...
+ */
+export interface NetworkPolicyPortProps {
+
+  /**
+   * Specific port number.
+   */
+  readonly port?: number;
+
+  /**
+   * End port (relative to `port`). Only applies if `port` is defined.
```
iliapolo, line 22: Adds that this creates a range

src/network-policy.ts
```diff
@@ -0,0 +1,517 @@
...
+  /**
+   * SCTP.
+   */
+  SCTP = 'SCTP',
+}
+
+/**
+ * Default behaviors of network traffic in policies.
+ */
+export enum NetworkPolicyTrafficDefault {
```
iliapolo, line 270: Add this to the docs...

src/network-policy.ts
```diff
@@ -0,0 +1,517 @@
...
+  /**
+   * The policy denies all traffic.
+   * Since rules are additive, additional rules or policies can allow
+   * specific traffic.
+   */
+  DENY = 'DENY',
```
iliapolo, line 276: Consider creating specific constructs for these:

- `AllowAllEgressNetworkPolicy`...
- `AllowAllIngressNetworkPolicy`...

src/network-policy.ts
iliapolo: Added

src/network-policy.ts
iliapolo: Added

src/network-policy.ts
iliapolo: Maybe we will add those in the future, if we see some confusion arising around it. 
These new constructs will then simply use the current API.

src/network-policy.ts
```diff
@@ -0,0 +1,522 @@
...
+  public static ipv4(cidrIp: string, except?: string[]): NetworkPolicyIpBlock {
+    const cidrMatch = cidrIp.match(/^(\d{1,3}\.){3}\d{1,3}(\/\d+)?$/);
+
+    if (!cidrMatch) {
+      throw new Error(`Invalid IPv4 CIDR: "${cidrIp}"`);
+    }
+
+    if (!cidrMatch[2]) {
+      throw new Error(`CIDR mask is missing in IPv4: "${cidrIp}". Did you mean "${cidrIp}/32"?`);
+    }
```
Chriscbr, line 176: Nice error handling

src/network-policy.ts
```diff
@@ -0,0 +1,522 @@
...
+}
+
+/**
+ * Describes how the network policy should configure egress / ingress traffic.
+ */
+export interface NetworkPolicyTraffic {
+
+  /**
+   * Specifies the default behavior of the policy when
+   * no egress rules are defined.
```
Chriscbr: ?
```suggestion
   * no egress or ingress rules are defined.
```

src/network-policy.ts
```diff
@@ -0,0 +1,522 @@
...
+   * Since rules are additive, additional rules or policies can allow
+   * specific traffic.
+   */
+  DENY = 'DENY',
+  /**
+   * The policy denies all traffic (either ingress or egress).
+   * Since rules are additive, no additional rule or policies can
+   * subsequently deny the traffic.
+   */
+  ALLOW = 'ALLOW',
```
Chriscbr: Is this a typo (should it say "The policy allows all traffic")?

src/network-policy.ts
```diff
@@ -0,0 +1,522 @@
...
+/**
+ * Properties for `NetworkPolicy`.
+ */
+export interface NetworkPolicyProps extends base.ResourceProps {
+
+  /**
+   * Which pods does this policy object applies to.
+   *
+   * This can either be a single pod / workload, or a grouping of pods selected
+   * via the `Pod.select` function. Rules is applied to any pods selected by this property.
```
Chriscbr:
```suggestion
   * via the `Pods.select` function. Rules is applied to any pods selected by this property.
```

src/pod.ts
```diff
@@ -1471,3 +1517,233 @@ export class PodScheduling {
...
+/**
+ * Isolation determines which policies are created
+ * when allowing connections from a pod to peers.
+ */
+export enum PodConnectionsIsolation {
+
+  /**
+   * Only creates network policies that select the pod.
+   */
+  POD = 'POD',
```
Chriscbr: Suggestion
```suggestion
/**
 * Isolation determines which policies are created
 * when allowing connections from a pod to peers
 * or from a pod workload to peers.
 */
export enum PodConnectionsIsolation {

  /**
   * Only creates network policies that select the pod or workload.
   */
  POD = 'POD',
```

src/pod.ts
```diff
@@ -1471,3 +1517,233 @@ export class PodScheduling {
...
+   *
+   * @example
+   *
+   * // create only an egress policy that selects the 'web' pod to allow outgoing traffic
+   * // to the 'redis' pod. this requires the 'redis' pod to not be isolated for ingress.
+   * web.connections.allowTo(redis, { isolation: Isolation.POD })
+   *
+   * // create only an ingress policy that selects the 'redis' peer to allow incoming traffic
+   * // from the 'web' pod. this requires the 'web' pod to not be isolated for egress.
+   * web.connections.allowTo(redis, { isolation: Isolation.PEER })
```
Chriscbr, line 1614: Nice examples! 👍

src/pod.ts
```diff
@@ -1471,3 +1517,233 @@ export class PodScheduling {
...
+
+    return ports;
+  }
+
+  private peerAddress(config: networkpolicy.NetworkPolicyPeerConfig): string {
+    const md5 = crypto.createHash('md5');
+    const data = config.ipBlock ? this.stringifyIpBlock(config.ipBlock) : this.stringifyPodSelector(config.podSelector!);
+    md5.update(data);
+    return md5.digest('hex');
+  }
```
Chriscbr: Even though this is private, it would be nice to have a comment explaining why this function exists / what purpose it's serving for future maintainers 🙂

Chriscbr: Also out of curiosity how long is the address this generates? If it's long we might want to slice of the last N characters maybe

src/pod.ts
```diff
@@ -1471,3 +1517,233 @@ export class PodScheduling {
...
+  private allow(direction: 'Ingress' | 'Egress', peer: networkpolicy.INetworkPolicyPeer, options: PodConnectionsAllowToOptions | PodConnectionsAllowFromOptions = {}) {
+
+    const config = peer.toNetworkPolicyPeerConfig();
+    networkpolicy.validatePeerConfig(config);
+
+    const peerAddress = this.peerAddress(config);
+
+    if (!options.isolation || options.isolation === PodConnectionsIsolation.POD) {
+
+      const src = new networkpolicy.NetworkPolicy(this.instance, `Allow${direction}${peerAddress}`, {
```
Chriscbr, line 1652: Can we add unit tests for this, checking that you can call allowTo twice (on different inputs), and that calling allowTo twice with the same input does something reasonable? (Same for allowFrom probably)

test/pod.test.ts
```diff
@@ -762,3 +762,508 @@ test('can select pods', () => {
...
+
+    const chart = Testing.chart();
+    const pod = new kplus.Pod(chart, 'Pod', {
+      containers: [{ image: 'pod' }],
+    });
+
+    pod.connections.allowFrom(kplus.NetworkPolicyIpBlock.anyIpv4());
+    expect(Testing.synth(chart)).toMatchSnapshot();
+
+  });
```
Chriscbr: (optional) Any way to dedup these tests by running the same subset of scenarios for `allowTo` and `allowFrom` maybe? `test.each` might help here

All that said, big kudos for testing all of these cases, I definitely wouldn't have caught all of them myself 😃

test/__snapshots__/pod.test.ts.snap
```diff
@@ -24,6 +24,2904 @@ Object {
...
+      },
+      "setHostnameAsFQDN": false,
+    },
+  },
+  Object {
+    "apiVersion": "networking.k8s.io/v1",
+    "kind": "NetworkPolicy",
+    "metadata": Object {
+      "name": "p-allowegressc88db22008a01c8e2b472a6a5ce6991f-c8d105cb",
+      "namespace": "default",
```
Chriscbr: This seems a little weird. NetworkPolicy defaults to the same namespace as the pod it's selecting (as required) -- but in this case, the original pod wasn't defined with a namespace, so is it possible the network policy should also be defined without a namespace (instead of "default")?

I wonder if this would allow users to use their generated template in any namespace through `kubectl apply -f dist/ -n whatever`? I haven't used the kubctl API in a while so I might be totally off on this

src/pod.ts
```diff
@@ -1471,3 +1517,233 @@ export class PodScheduling {
...
+
+        // if the peer doesn't specify namespaces, we assume the same namespace.
+        namespaces = [this.instance.metadata.namespace ?? 'default'];
+
+      } else {
+
+        // a peer cannot specify namespaces by labels because
+        // we won't be able to extract the names of those namespaces.
+        if (podSelectorConfig.namespaces.labelSelector && !podSelectorConfig.namespaces.labelSelector.isEmpty()) {
+          throw new Error('Unable to create peer policy. Peer must specify namespaces only by name');
```
Chriscbr: Could we provide the name of the pod / workflow / selector  that we're creating a peer policy for in the error messages?

src/pod.ts
```diff
@@ -1471,3 +1517,233 @@ export class PodScheduling {
...
+          throw new Error('Unable to create peer policy. Peer must specify namespace names');
+        }
+
+        namespaces = podSelectorConfig.namespaces.names;
+      }
+
+      for (const name of namespaces) {
+        switch (direction) {
+          case 'Egress':
+            new networkpolicy.NetworkPolicy(this.instance, `AllowIngress${name}${peerAddress}`, {
```
Chriscbr, line 1710: Just double checking, these are meant to be opposites yes? (Egress and AllowIngress)

src/pod.ts
```diff
@@ -1471,3 +1517,233 @@ export class PodScheduling {
...
+              metadata: { namespace: name },
+              ingress: { rules: [{ peer: this.instance, ports: options.ports }] },
+            });
+            break;
+          case 'Ingress':
+            new networkpolicy.NetworkPolicy(this.instance, `AllowEgress${name}${peerAddress}`, {
+              selector: { toPodSelectorConfig: () => podSelectorConfig },
+              metadata: { namespace: name },
+              egress: { rules: [{ peer: this.instance, ports: options.ports }] },
+            });
```
Chriscbr: Code between these cases looks similar, maybe it could instead be 
```
new networkpolicy.NetworkPolicy(this.instance, `Allow${oppositeDirection}${name}${peerAddress}`, {
  selector: { toPodSelectorConfig: () => podSelectorConfig },
  metadata: { namespace: name },
  egress: { rules: [{ peer: this.instance, ports: options.ports }] },
});
```

src/network-policy.ts
```diff
@@ -0,0 +1,522 @@
...
+    super(scope, id);
+
+    this._podSelectorConfig = (props.selector ?? pod.Pods.all()).toPodSelectorConfig();
+
+    let ns;
+
+    if (!props.metadata?.namespace) {
+
+      if (this._podSelectorConfig.namespaces?.labelSelector && !this._podSelectorConfig.namespaces?.labelSelector.isEmpty()) {
+        throw new Error('Unable to create a network policy for a selector that selects pods in namespaces based on labes');
```
Chriscbr:
```suggestion
        throw new Error('Unable to create a network policy for a selector that selects pods in namespaces based on labels');
```

src/network-policy.ts
```diff
@@ -0,0 +1,522 @@
...
+      }
+
+      if (this._podSelectorConfig.namespaces?.names && this._podSelectorConfig.namespaces.names.length > 1) {
+        throw new Error('Unable to create a network policy for a selector that selects pods in multiple namespace');
```
Chriscbr:
```suggestion
        throw new Error('Unable to create a network policy for a selector that selects pods in multiple namespaces');
```

Chriscbr (approved):
LGTM. Mostly nitpicks

src/network-policy.ts
iliapolo: Typo 👍

iliapolo: Typo 👍

src/pod.ts
iliapolo: Removed this altogether

src/pod.ts
iliapolo: Adde a unit test for different inputs. As far as the same input, it will currently throw a "Construct already exists" error, which I know isn't great, but we can address this later if needed.

test/pod.test.ts
iliapolo: I really don't like the `test.each` because it makes it harder to debug...in any case, maybe later :)

src/pod.ts
iliapolo: Yeap

test/__snapshots__/pod.test.ts.snap
iliapolo: Great call out, yeah this was a quirk from a previous poorly modeled design - removed it so it now just leaves this blank 👍

src/pod.ts
iliapolo: Now that peers are constructs, they have id's :) - improved this to include both peer and pod ids.

src/pod.ts
iliapolo: Not so simple, notice the `egress / ingress` property need to alternate as well.

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-22/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/753)|
|✅|k8s-23/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/754)|

### Questions ?

PR #741, opened 2022-05-23
fix(container): unable to set partial container resources

nielsslot:
I noticed I couldn't specify just a limit or request in the container resources. I could either fully specify the limits and requests for both cpu and memory, or could I choose not specify anything at all, but nothing in between. After this change I can, for example, specify only the request and limits for memory, leaving cpu to best effort.

I consider this a fix since the code was already checking for non-null values in the `ContainerResources` parameter (see container.ts, lines 683-686). The values couldn't be null before (they were marked required), so these checks weren't functional. With this fix they can be null and the checks are functional again.


iliapolo (approved):
Thanks!

PR #755, opened 2022-05-25
feat(service): select pods by selector

iliapolo:
Now that we formalized pod selection, let's use this in the `Service` resource to allow for standard selection. 

In addition, this PR removes the `addDeployment` method because it was pretty awkward to use. You could only invoke it once, because indeed a service cannot serve multiple deployments. 

Instead, you can now pass a `selector` at construction time, and optionally add additional ones post construction. 
Adding multiple selectors acts as an AND operator on the labels.

I also changed `service.serve` to `service.bind` since you don't serve ports, you bind to it. 

BREAKING CHANGE: `service.addDeployment` was removed. Use `selector` instead. Also, `service.serve` is renamed to `service.bind`.


test/service.test.ts
```diff
@@ -86,93 +82,27 @@ test('Can serve by port', () => {
...
-
-  const chart = Testing.chart();
-
-  const service = new kplus.Service(chart, 'service');
-  const deployment = new kplus.Deployment(chart, 'dep');
-  deployment.addContainer({ image: 'foo', port: 7777 });
-
-  service.addDeployment(deployment);
-
-  const expectedSelector = { 'cdk8s.io/metadata.addr': 'test-dep-c8cc9f8f' };
```
Chriscbr, line 127: Is there still a way to achieve this? It would be nice to have a test for that, even if it's not explicitly using a hand-crafted method (`addDeployment`)

Chriscbr (approved):
LGTM! Only feedback is that if we're removing `addDeployment`, it might help to add a small code snippet in the PR description showing how that behavior might be achieved now. (Or maybe recommend using `Deployment.exposeViaService`)

test/service.test.ts
iliapolo: Yeah I added a test, you basically just pass the deployment object to the `selector` property when creating the service, or in `service.select`.

PR #756, opened 2022-05-25
feat: resource grant methods

iliapolo:
Add a `permissions` API on resources, that exposes `grant` methods on subjects.


src/pod.ts
```diff
@@ -143,6 +144,27 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
     this._volumes.set(vol.name, vol);
   }
 
+  /**
+   * @see ISubect.toSubjectConfiguration()
```
Chriscbr, line 148:
```suggestion
   * @see ISubject.toSubjectConfiguration()
```

src/role-binding.ts
```diff
@@ -202,6 +216,17 @@ export class User extends Construct implements ISubject {
     super(scope, id);
     this.name = name;
   }
+
+  /**
+   * @see ISubect.toSubjectConfiguration()
```
Chriscbr, line 221:
```suggestion
   * @see ISubject.toSubjectConfiguration()
```

Chriscbr (approved):

PR #841, opened 2022-06-15
fix(pod): unable to use `LabelSelector` class in python

### LINKED ISSUE #745: Python, deployment select issue
👋 

I am using the python client with version `2.0.0b15`, and I am having issues when it comes to the select command on the deployment object.

code:
```python
deployment.select(LabelSelector.of({"labels": {"app": "my-app"}}))
```

error message
```
raise JSIIError(resp.error) from JavaScriptError(resp.stack)
jsii.errors.JSIIError: Expected object reference, got {"$jsii.map":{"labels":{"$jsii.map":{"app":"my-app"}}}}
```

I can't really find any documentation for this (as the documentation is not up to date), so I am trying to read the source code after I downloaded the package from pypi, and the tests written in typescript.

Approach from the tests
```js
  const expectedSelector = { foo: 'bar' };

  deployment.select(kplus.LabelSelector.of({ labels: { foo: expectedSelector.foo } }));
```

Any indications of what I am doing wrong here?

This is how the jsii python code looks for the package:
```python
class LabelSelector(metaclass=jsii.JSIIMeta, jsii_type="cdk8s-plus-22.LabelSelector"):
    '''Match a resource by labels.'''

    @jsii.member(jsii_name="of") # type: ignore[misc]
    @builtins.classmethod
    def of(
        cls,
        options: typing.Optional["LabelSelectorOptions"] = None,
    ) -> "LabelSelector":
        '''
        :param options: -
        '''
        return typing.cast("LabelSelector", jsii.sinvoke(cls, "of", [options]))
```

jaeheonji:
Fixes #745 

Hi, cdk8s-team! Thanks for maintaining such a awesome project.
Recently, I'm testing `cdk8s-plus` based on python at work. but, As with #745, currently specifying a selector for deployment in python does not work.

Because `LabelsSelectorOption` is `class`, I could not assign values to `labels` and `expressions` properties (readonly).
Perhaps there is a problem after converting to python code by `jsii`.

### DIFF
--- a/src/pod.ts
+++ b/src/pod.ts
@@ -423,7 +423,7 @@ export interface PodProps extends AbstractPodProps {}
 /**
  * Options for `LabelSelector.of`.
  */
-export class LabelSelectorOptions {
+export interface LabelSelectorOptions {
 
   /**
    * Strict label matchers.


iliapolo (approved):
Oh good catch, thanks @jaeheonji !

PR #987, opened 2022-07-27
feat: Add support for configuring nodeSelector

### LINKED ISSUE #986: Add support for configuring nodeSelector
Here is the k8s documentation https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector

rvshettigar:
Signed-off-by: rakshith <rakshith@yahooinc.com>

Fixes #986

### DIFF
--- a/src/pod.ts
+++ b/src/pod.ts
@@ -19,6 +19,7 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
   public readonly dns: PodDns;
   public readonly dockerRegistryAuth?: secret.DockerConfigSecret;
   public readonly automountServiceAccountToken: boolean;
+  public readonly nodeSelector?: { [key: string]: string };
 
   private readonly _containers: container.Container[] = [];
   private readonly _initContainers: container.Container[] = [];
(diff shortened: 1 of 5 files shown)


iliapolo:
@rvshettigar Thanks for taking the time on this, but there is actually already a way to configure node selectors, its just a little different than the core resource API.

See https://cdk8s.io/docs/latest/plus/pod/#node-selection

Let me know if this fully addresses your issue.

rvshettigar:
Thank you for pointing it out. I tried out what you suggested and it generates the following in the manifest
```
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: memory
                    operator: In
                    values:
                      - high
  ```
It is different than the one generated by `nodeSelector`, but if it has the same effect then we don't need this pull request.
I guess there is also the option of using escape hatch, since manifest with `nodeSelector` is a bit easier to understand.

iliapolo:
@rvshettigar Yeah I agree the manifest itself is a little less clear, but that's why we have an intent based API for it :)

According to k8s [docs](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity) the effect is the same.

Closing this PR 👍

PR #1052, opened 2022-08-17
feat(cronjob): introduce `CronJob` construct

### LINKED ISSUE #376: `CronJob` L2
https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

### LINKED ISSUE #813: Configure default deadline for CronJob
In `CronJob`, `startingDeadlineSeconds` should default to something reasonable. 

### Additional resources

- https://hub.datree.io/built-in-rules/ensure-cronjob-deadline

### LINKED ISSUE #814: Configure default concurrency policy for `CronJob`
Should be set to `Forbid`.

### Additional resources

- https://hub.datree.io/built-in-rules/prevent-cronjob-concurrency

vinayak-kukreja:
This PR adds the Kubernetes [CronJob Resource L2](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/cron-job-v1/) to cdk8s-plus. This is part of the effort to make cdk8s+ GA. 

Having this L2 abstraction would enable the customer to provision Kubernetes CronJob Resource using cdk8s+ and that would generate/add it to the Kubernetes manifest.

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/376
Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/813
Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/814

## Next Steps

Since timezone is behind a feature gate and works with K8s-24, we would need to fix the backport PR.

## DCO


src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+import * as timezone from 'moment-timezone';
+import * as k8s from './imports/k8s';
+import { JobProps } from './job';
+import { RestartPolicy } from './pod';
+import * as workload from './workload';
+
+/**
+ * Concurrency policy for CronJobs.
+ */
+export enum ConcurrencyPolicy {
```
iliapolo, line 11: Add docstrings to each enum value, explaining what it does / means.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Specifies the timezone for the job. This helps aligining the schedule to follow the specified timezone.
+   *
+   * @default - Timezone of kube-controller-manager process.
+   */
+  readonly timeZone?: string;
+
+  /**
+   * Specifies the concurrency policy for the job.
+   *
+   * @default - Allow
```
iliapolo: As mentioned [here](https://github.com/cdk8s-team/cdk8s-plus/issues/814), this should be `Forbid`.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Properties of the Job that is being scheduled by CronJob.
+   */
+  readonly jobProperties: JobProps;
+
+  /**
+   * Specifies the time in which the job would run again.
+   *
+   * @default - Runs at every minute.
+   */
+  readonly schedule?: CronOptions;
```
iliapolo: I believe this should be typed as `Schedule`. `CronOptions` are only needed when configuring `Schedule.cron` as the value.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/**
+ * Properties for `CronJob`.
+ */
+export interface CronJobProps extends workload.WorkloadProps {
+  /**
+   * Properties of the Job that is being scheduled by CronJob.
+   */
+  readonly jobProperties: JobProps;
```
iliapolo: Because `JobProps` also extends `WorkloadProps`, this makes it so pod properties can be specified twice. For example:

```ts
new CronJob(this, 'CronJob', {
  jobProperties: {
    containers: ...,
  },
  containers: ...,
});
```

What we need here is actually:

```suggestion
export interface CronJobProps extends JobProps {
```

This would make the API look like:

```ts
new CronJob(this, 'CronJob', {
  containers: ...,
});
```

Which is consistent with what we have for other workload resources. The only potential problem here is with collisions, but since we control the property names, we can always workaround that if needed.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * Specifies the time in which the job would run again.
+   *
+   * @default - Runs at every minute.
```
iliapolo: A default value here is somewhat surprising IMO. There's no sane default we can apply because we don't know the nature of the task, let's make this required.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+  /**
+   * Specifies the concurrency policy for the job.
+   *
+   * @default - Allow
+   */
+  readonly concurrencyPolicy?: ConcurrencyPolicy;
+
+  /**
+   * Specifies the time by which a recurring job should be scheduled after it has passed its expected scheduling time.
+   * Missed jobs are counted as failed ones.
```
iliapolo: Don't be shy putting as much information as needed here to make things clear. This prevents the user from having to exit their IDE, which is always unpleasant.

  ```suggestion
   * Kubernetes attempts to start cron jobs at its schedule time, but this is not guaranteed. This deadline specifies
   * how much time can pass after a schedule point, for which kubernetes can still start the job.
   * For example, if this is set to 100 seconds, kubernetes is allowed to start the job at a maximum 100 seconds after
   * the scheduled time.
   *
   * Note that the Kubernetes CronJobController checks for things every 10 seconds, for this reason, a deadline below 10
   * seconds is not allowed, as it may cause your job to never be scheduled.
   *
   * In addition, kubernetes will stop scheduling jobs if more than 100 schedules were missed (for any reason).
   * This property also controls what time interval should kubernetes consider when counting for missed schedules.
   *
   * For example, suppose a CronJob is set to schedule a new Job every one minute beginning at 08:30:00,
   * and its `startingDeadline` field is not set. If the CronJob controller happens to be down from 08:29:00 to 10:21:00,
   * the job will not start as the number of missed jobs which missed their schedule is greater than 100.
   * However, if `startingDeadline` is set to 200 seconds, kubernetes will only count 3 missed schedules, and thus
   * start a new execution at 10:22:00.
```

I hope I understood the meaning of this property, let me know if I got anything wrong. Note that the example is taken directly from k8s [docs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations).

File: src/cron-job.ts [hunk 7]
```diff
@@ -0,0 +1,251 @@
...
+   *
+   * @default - There is no deadline.
+   */
+  readonly startingDeadline?: Duration;
+
+  /**
+   * Specifies if the recurring job should be suspended.
```
iliapolo:
```suggestion
   * Specifies if the recurring job should be suspended. Only applies to future executions, current ones are remained untouched. 
```

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   *
+   * @default - false.
```
iliapolo: We use `-` when the default describes a behavior, if its a literal value, we do:

```suggestion
   * @default false
```

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  readonly suspend?: boolean;
+
+  /**
+   * Specifies the number of successful jobs history retained.
+   *
+   * @default - 3.
```
iliapolo:
```suggestion
   * @default 3
```

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  readonly successfulJobsRetained?: number;
+
+  /**
+   * Specifies the number of failed jobs history retained.
+   *
+   * @default - 1.
```
iliapolo:
```suggestion
   * @default 1
```

src/cron-job.ts
iliapolo: Does this actually retain the `Job` resource that kubernetes creates? i.e if you run `kubectl get jobs` you'll see them? If so, make this clear.

src/cron-job.ts
iliapolo: Same as above

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+ * A CronJob is responsible for creating a Job and scheduling it based on provided scheduled. This helps running Jobs in a recurring manner.
+ */
+export class CronJob extends workload.Workload {
+  /**
+   * Properties of the recurring `Job`.
+   */
+  public readonly jobProperties: JobProps;
+
+  /**
+   * Schedule for which to run the recurring job. By default, the job would be scheduled to run every minute.
```
iliapolo: These are now no longer properties, they are attributes / fields. That is, they should describe the behavior of a would be instance, but the behavior of the current one.

```suggestion
   * The schedule this cron job is scheduled to run in.
```

Please rephrase all of these to convey the correct context.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  public readonly concurrencyPolicy?: string;
+
+  /**
+   * Deadline to start next instance of job after the expected scheduling time. The job is considered as failed if it misses this deadline.
+   */
+  public readonly startingDeadline?: Duration;
+
+  /**
+   * Flag that tells to suspend the upcoming executions of the job.
```
iliapolo: Another example:

```suggestion
   * Whether or not the cron job is currently suspended or not.
```

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  public readonly suspend?: boolean;
+
+  /**
+   * The number of successful jobs to retain.
+   */
+  public readonly successfulJobsRetained?: number;
```
iliapolo: Since this has a default value, it shouldn't be possible for it to be `undefined`.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+  protected readonly apiObject: ApiObject;
+
+  /**
+   * Represents the resource type.
+   */
+  public readonly resourceType = 'cronjob';
+
+  constructor(scope: Construct, id: string, props: CronJobProps) {
+    super(scope, id, {
+      restartPolicy: RestartPolicy.ON_FAILURE,
```
iliapolo: Why did you choose this default?

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+      select: false, //Why is this false in Job?
```
iliapolo: Its because jobs are not required to match labels and selectors with its pod definitions.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+    if (!this.isValidTimezone(props.timeZone)) {
+      throw new Error(`Invalid timezone: ${props.timeZone}`);
+    }
+
+    this.jobProperties = props.jobProperties;
+    this.schedule = props.schedule;
+    this.timeZone = props.timeZone;
+    this.concurrencyPolicy = props.concurrencyPolicy;
+    this.startingDeadline = props.startingDeadline;
+    this.suspend = props.suspend;
```
iliapolo: This is where the default value would go, which will remove the need for the optionality of the attribute.

```suggestion
    this.suspend = props.suspend ?? false;
```

Even though it's a kubernetes default, we commonly hoist it to the manifest. It lets users see more clearly what happens in the manifest.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * @internal
+   */
+  public _toKube(): k8s.CronJobSpec {
+    return {
+      concurrencyPolicy: this.concurrencyPolicy ?? ConcurrencyPolicy.FORBID,
+      failedJobsHistoryLimit: this.failedJobsRetained,
+      jobTemplate: {
+        metadata: this.jobProperties.metadata,
```
iliapolo: You'll need to accept a new `jobMetadata` property and use it here after you apply the props hierarchy change. This is ok, the construct will have 3 metadata objects it can configure:

- `metadata`: The metadata of the `CronJob` resource.
- `jobMetadata`: The metadata of the `Job` resource that `CronJob` creates on our behalf.
- `podMetadata`: The metadata of the `Pod` resource that `Job` creates on our behalf.

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+      successfulJobsHistoryLimit: this.successfulJobsRetained,
+      suspend: this.suspend,
+      timeZone: this.timeZone,
+    };
+  }
+
+  /**
+   * Returns the job spec.
+   */
+  private getJobSpec(): k8s.JobSpec {
```
iliapolo: We usually refrain from using `get` for anything other than a getter (which is just direct property access in TS).
Also, start these conversion methods with `_`.  

Just some conventions we gathered along the way :)

```suggestion
  private _toJobSpec(): k8s.JobSpec {
```

src/cron-job.ts
```diff
@@ -0,0 +1,251 @@
...
+   */
+  public _toKube(): k8s.CronJobSpec {
+    return {
+      concurrencyPolicy: this.concurrencyPolicy ?? ConcurrencyPolicy.FORBID,
+      failedJobsHistoryLimit: this.failedJobsRetained,
+      jobTemplate: {
+        metadata: this.jobProperties.metadata,
+        spec: this.getJobSpec(),
+      },
+      schedule: this.getCronExpression(this.schedule),
```
iliapolo: The `getCronExpression` method shouldn't exist. This should just be:

```suggestion
      schedule: this.schedule.toCronExpression(),
```

And we should add `toCronExpression` on the [`Schedule`](https://github.com/cdk8s-team/cdk8s-core/blob/2.x/src/schedule.ts) class. Also looking at that class, I think we got it wrong because it provide functionality not available in Kubernetes (like rates). We need to first fix it.

test/cron-job.test.ts
```diff
@@ -0,0 +1,296 @@
...
+  expect(spec.successfulJobsHistoryLimit).toEqual(undefined);
+  expect(spec.failedJobsHistoryLimit).toEqual(undefined);
+
+  expect(spec.jobTemplate.spec.activeDeadlineSeconds).toEqual(20);
+  expect(spec.jobTemplate.spec.backoffLimit).toEqual(4);
+  expect(spec.jobTemplate.spec.ttlSecondsAfterFinished).toEqual(1);
+  expect(spec.jobTemplate.spec.template.spec.containers[0].image).toEqual('image');
+});
+
+test('allow setting schedule using cron options', () => {
```
iliapolo: No need for a different test case for each property. Just have one test that doesn't set any custom value, and one test that sets all available custom values. Then snapshot test it.

iliapolo (changes requested):

src/cron-job.ts
vinayak-kukreja: Will add it.

src/cron-job.ts
vinayak-kukreja: Right, will update this.

src/cron-job.ts
vinayak-kukreja: Ahh, I see, will update it.

src/cron-job.ts
vinayak-kukreja: Yeah, I was on the fence too to give this a default. Will update this to required.

src/cron-job.ts
vinayak-kukreja: Makes sense. Will update it. :)

I was also not sure if we can use examples from official documentation?

src/cron-job.ts
vinayak-kukreja: Will update it.

src/cron-job.ts
vinayak-kukreja: Ok, will remember this. Thank you :)

src/cron-job.ts
vinayak-kukreja: Will update it.

src/cron-job.ts
vinayak-kukreja: Will update it.

src/cron-job.ts
vinayak-kukreja: I see, will update it.

src/cron-job.ts
vinayak-kukreja: Will update it.

src/cron-job.ts
vinayak-kukreja: I see, ok, I thought the selector would need them to be the same.

src/cron-job.ts
vinayak-kukreja: Makes sense, will update it.

src/cron-job.ts
vinayak-kukreja: Will update it, thank you for letting me know about these conventions :)

src/cron-job.ts
vinayak-kukreja: Yes, on it.

test/cron-job.test.ts
vinayak-kukreja: Great, will remove these.

src/cron-job.ts
vinayak-kukreja: Hmm, its default is applied by Kubernetes that's why I had this as undefined.

src/cron-job.ts
vinayak-kukreja: I imagined it would help with job completion but now that I think about it, even if this restarts, I am not sure the cronjob would still keep track of this particular completion and would probably mark it as failed or reschedule it.

src/cron-job.ts
vinayak-kukreja: Makes sense, will update it.

vinayak-kukreja: Hey Eli, I see we can accept metadata in JobProperties right now. Is that what you are referring to or I need to introduce `jobMetadata` within `jobProperties`?

iliapolo: So `metadata` and `podMetadata` will be automatically available by extending `JobProperties`. You only need to add `jobMetadata`.

iliapolo:
@vinayak-kukreja Let me know when you ready for another review, just re-request the review on the top right.

package.json
```diff
@@ -54,7 +54,7 @@
     "@types/node": "^14",
     "@typescript-eslint/eslint-plugin": "^5",
     "@typescript-eslint/parser": "^5",
-    "cdk8s": "2.4.10",
+    "cdk8s": "2.4.16",
```
vinayak-kukreja: This manual update is so the new Cron Class is accessible for CronJob construct.

src/cron-job.ts
vinayak-kukreja: Hmm, I am still not sure if I understand. Will sync with you tomorrow about this.

From my understanding, `this.jobProperties.metadata` is the `jobMetadata` and this can right now be passed in JobProperties.

vinayak-kukreja: And looks like the `k8s.CronJobSpec` does not have a metadata field like the `JobSpec` 🤔

src/cron-job.ts
```diff
@@ -0,0 +1,227 @@
...
+}
+
+/**
+ * Properties for `CronJob`.
+ */
+export interface CronJobProps extends JobProps {
+  /**
+   * Metadata of the Job that is being scheduled by CronJob.
+   */
+  readonly jobMetadata?: ApiObjectMetadata;
```
vinayak-kukreja: Adding a custom name here has no effect. CronJob decides the name of the job being scheduled on its own.

src/cron-job.ts
```diff
@@ -0,0 +1,227 @@
...
+   * @see {@link https://en.wikipedia.org/wiki/List_of_tz_database_time_zones} for list of valid timezone values.
+   *
+   * @default - Timezone of kube-controller-manager process.
+   */
+  readonly timeZone?: string;
+
+  /**
+   * Specifies the concurrency policy for the job.
+   *
+   * @default Forbid
```
iliapolo:
```suggestion
   * @default ConcurrencyPolicy.Forbid
```

src/cron-job.ts
```diff
@@ -0,0 +1,227 @@
...
+   * In addition, kubernetes will stop scheduling jobs if more than 100 schedules were missed (for any reason).
+   * This property also controls what time interval should kubernetes consider when counting for missed schedules.
+   *
+   * For example, suppose a CronJob is set to schedule a new Job every one minute beginning at 08:30:00,
+   * and its `startingDeadline` field is not set. If the CronJob controller happens to be down from 08:29:00 to 10:21:00,
+   * the job will not start as the number of missed jobs which missed their schedule is greater than 100.
+   * However, if `startingDeadline` is set to 200 seconds, kubernetes will only count 3 missed schedules, and thus
+   * start a new execution at 10:22:00.
+   *
+   * @default - There is no deadline.
```
iliapolo: I think you forgot to update this after adding a default value of 10 seconds for this property.

src/cron-job.ts
```diff
@@ -0,0 +1,227 @@
...
+   */
+  readonly startingDeadline?: Duration;
+
+  /**
+   * Specifies if the recurring job should be suspended. Only applies to future executions, current ones are remained untouched.
```
iliapolo: *"recurring job"* is a little confusing as it sounds like you're referring to some internal job or something. 

```suggestion
   * Specifies if the cron job should be suspended. Only applies to future executions, current ones are remained untouched.
```

src/cron-job.ts
```diff
@@ -0,0 +1,227 @@
...
+      ...props,
+    });
+
+    this.apiObject = new k8s.KubeCronJob(this, 'Resource', {
+      metadata: props.metadata,
+      spec: Lazy.any({ produce: () => this._toKube() }),
+    });
+
+    if (props.startingDeadline != undefined && props.startingDeadline.toSeconds() < 10) {
+      throw new Error(`The starting deadline cannot be less than 10 seconds since the Kubernetes CronJobController checks things every 10 seconds and hence the CronJob may not be scheduled. The value passed is: ${props.startingDeadline}`);
```
iliapolo: Try using the same names as the API, so its clear what needs to be changed. 

```suggestion
      throw new Error(`The 'startingDeadline' property cannot be less than 10 seconds since the Kubernetes CronJobController checks things every 10 seconds and hence the CronJob may not be scheduled. The value passed is: ${props.startingDeadline}`);
```

src/cron-job.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    if (props.ttlAfterFinished != undefined && (props.successfulJobsRetained != undefined || props.failedJobsRetained != undefined)) {
+      throw new Error('The ttlAfterFinished cannot be set if cron job is retaining successful or failed job runs. This would cause the retention of jobs to not work properly since it would delete the job based on its value.');
```
iliapolo:
```suggestion
      throw new Error('The 'ttlAfterFinished' property cannot be set if 'successfulJobsRetained' or 'failedJobsRetained' is set. This would cause the retention of jobs to not work properly since it would delete the job based on its value.');
```

src/cron-job.ts
iliapolo: Hmm. Then lets just remove this for now, I don't see a good use-case for explicitly setting this metadata right now. We can always add if needed.

test/cron-job.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  const schedule = new Cron(cronOptions);
+
+  const defaultChild = Node.of(new kplus.CronJob(chart, 'CronJob', {
+    schedule: schedule,
+  })).defaultChild as ApiObject;
+
+  expect(defaultChild.kind).toEqual('CronJob');
+});
+
+test('allow setting jobProperties / default configuration', () => {
```
iliapolo:
```suggestion
test('default configuration', () => {
```

test/cron-job.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  expect(spec.suspend).toEqual(false);
+  expect(spec.successfulJobsHistoryLimit).toEqual(3);
+  expect(spec.failedJobsHistoryLimit).toEqual(1);
+
+  expect(spec.jobTemplate.spec.activeDeadlineSeconds).toEqual(60);
+  expect(spec.jobTemplate.spec.backoffLimit).toEqual(4);
+  expect(spec.jobTemplate.spec.template.spec.containers[0].image).toEqual('image');
+});
+
+test('allows setting all options', () => {
```
iliapolo:
```suggestion
test('custom configuration', () => {
```

test/cron-job.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+test('allow setting jobProperties / default configuration', () => {
+
+  const chart = Testing.chart();
+
+  const schedule = Cron.everyMinute();
+
+  new kplus.CronJob(chart, 'CronJob', {
+    activeDeadline: Duration.seconds(60),
+    backoffLimit: 4,
```
iliapolo: No need. This test should just test the required properties and assert against the default values.

iliapolo (changes requested):

src/cron-job.ts
vinayak-kukreja: Will remove it.

src/cron-job.ts
vinayak-kukreja: Will update this.

src/cron-job.ts
vinayak-kukreja: Right, will update this.

src/cron-job.ts
vinayak-kukreja: Will update this.

src/cron-job.ts
vinayak-kukreja: Will update this.

src/cron-job.ts
vinayak-kukreja: Will update this.

test/cron-job.test.ts
vinayak-kukreja: Will update this.

test/cron-job.test.ts
vinayak-kukreja: Will update this.

test/cron-job.test.ts
vinayak-kukreja: Makes sense, will update this.

iliapolo (approved):

iliapolo:
@vinayak-kukreja As you predicted, the backport PR's are failing with:

```console
' is not assignable to type 'CronJobSpec'.
  Object literal may only specify known properties, and 'timeZone' does not exist in type 'CronJobSpec'.
```

I guess just remove the `timeZone` property?

PR #1069, opened 2022-08-22
feat(Container): add "addPort" method for Container class

### LINKED ISSUE #28: Enhanced port
Current containers can only export a single port. It has even limited set of features, no named port, cannot set protocols.

### Proposed Solution:
```ts
const deployment = new Deployment(...);
const container = deployment.addContainer({
  ...
  ports: [
    { name: "http", containerPort: 80, }
  ]
});

const httpPort = container.getPort("http"); // == container.getPort(80);
const webhookPort = container.addPort(8080, { name: "webhook" });

deployment.expose(80, {
  targetPort: httpPort,
});
deployment.expose(8080, {
  targetPort: "webhook",
});

```

`deployment.expose()` also should validate given `targetPort` is actually owned by any of its child containers.
Also it would be better to warn/error for shadowed ports (proactively at the time of `addPort`, or reactively at the time of `expose`) since Kubernetes doesn't validate uniqueness of ports/names in a pod (https://github.com/kubernetes/kubernetes/issues/54213)

### Other:

* [x] :warning: This feature might incur a breaking change

---

segal90:
Fixes #28

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -414,6 +414,39 @@ export interface ContainerLifecycle {
 }
 
 /**
+ * ContainerPort represents a network port in a single container.
+ */
+export interface ContainerPort {
+  /**
+   * Number of port to expose on the pod's IP address. This must be a valid port number, 0 < x < 65536.
+   */
+  readonly port: number;
+
+  /**
+   * Number of port to expose on the host. If specified, this must be a valid port number, 0 < x < 65536. If HostNetwork is specified, this must match ContainerPort. Most containers do not need this.
+   */
+  readonly hostPort?: number;
+
+  /**
+   * What host IP to bind the external port to.
+   */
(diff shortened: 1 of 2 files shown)


segal90:
Hi @iliapolo @ryparker @vinayak-kukreja ,
I implemented the linked feature, it would be beneficial to everyone if you can help me to merge back this change.
Thanks

iliapolo:
@segal90 Thanks for picking this up! I'm afraid though this issue is more involved. There are few places in the library that assume a single container port for some default behavior, the implementation you suggested doesn't address this, and i'm honestly not entirely sure yet how we want to approach it. 

This issue is on our current iteration plan, so we will be picking it up fairly soon. Closing this PR, but thanks again and stay tuned :)

PR #1082, opened 2022-08-29
feat(container): default container resource requirements

### LINKED ISSUE #807: Configure default container resource requirements
Should apply to:

- [Memory request](https://hub.datree.io/built-in-rules/ensure-memory-request)
- [Memory limit](https://hub.datree.io/built-in-rules/ensure-memory-limit)
- [CPU request](https://hub.datree.io/built-in-rules/ensure-cpu-request)
- [CPU limit](https://hub.datree.io/built-in-rules/ensure-cpu-limit)

### Additional resources

- https://polaris.docs.fairwinds.com/checks/efficiency/

vinayak-kukreja:
As we will allow customers to integrate with manifest validation plugins soon, we see that one of the recommendation for K8s manifest is to have defined container resource limits. This would safeguard the containers to not have undefined limits which could end up in Out Of Memory crashes or performance degradation due to CPU over utilization.

Picking up these defaults from what Lambda follows [here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-function-common.html), let me know if you believe these can be updated. 

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/807

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -623,7 +623,7 @@ export class Container {
         limit: Cpu.millis(1500),
       },
       memory: {
-        request: Size.mebibytes(1769),
+        request: Size.mebibytes(512),
         limit: Size.mebibytes(2048),
       },
     };
--- a/src/container.ts
+++ b/src/container.ts
@@ -1,4 +1,4 @@
-import { Size as container } from 'cdk8s';
+import { Size as container, Size } from 'cdk8s';
 import * as configmap from './config-map';
 import * as handler from './handler';
 import * as k8s from './imports/k8s';
@@ -609,6 +609,17 @@ export class Container {
       throw new Error('Attempted to construct a container from a Container object.');
     }
 
+    const defaultResourceSpec: ContainerResources = {
+      cpu: {
+        request: Cpu.millis(1000),
+        limit: Cpu.millis(1500),
+      },
+      memory: {
+        request: Size.mebibytes(1769),
+        limit: Size.mebibytes(2048),
+      },
+    };
+
     this.name = props.name ?? 'main';
     this.image = props.image;
     this.port = props.port;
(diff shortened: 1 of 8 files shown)


src/container.ts
```diff
@@ -609,6 +609,17 @@ export class Container {
       throw new Error('Attempted to construct a container from a Container object.');
     }
 
+    const defaultResourceSpec: ContainerResources = {
```
iliapolo, line 620: These values need to be documented in the `resources` property of `ContainerProps`.

iliapolo (changes requested):

src/container.ts
vinayak-kukreja: Will add them.

src/container.ts
```diff
@@ -609,6 +617,17 @@ export class Container {
...
+      cpu: {
+        request: Cpu.millis(1000),
+        limit: Cpu.millis(1500),
+      },
+      memory: {
+        request: Size.mebibytes(1769),
```
iliapolo: I've thought about this more and I think this might be a bit much for a default. It will make it hard to deploy applications to a local cluster.

I think that the separation between `request` and `limit` gives us the ability to provide a good balance between developer friendly, and safe.

```suggestion
        request: Size.mebibytes(512),
```

iliapolo (changes requested):

src/container.ts
vinayak-kukreja: I see, will make the update. Do you want to reduce the CPU too for request?

iliapolo (approved):

src/container.ts
iliapolo: Nah. CPU is ok.

antholeole:
how do we unset these limits should we want to?

EDIT: nvm figured it out. just set the object to {}.

PR #1093, opened 2022-09-02
feat(container): default container startup probe

### LINKED ISSUE #808: Configure default readiness probe for containers
In case a container defines a port, we can default a readiness probe that knocks on that port via `Probe.fromTcpSocket`.
Also consider defaulting the liveness probe to this.

### Additional resources

- https://hub.datree.io/built-in-rules/ensure-readiness-probe
- https://polaris.docs.fairwinds.com/checks/reliability/

vinayak-kukreja:
Startup probe help Kubernetes determine if the container is ready and no other probe is initiated till this check is complete. Currently we leave it as undefined in the manifest if the values are not provided.
Now, if a port number is provided by the user, that will be knocked on to understand status of the container.

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/808


src/container.ts
```diff
@@ -609,13 +613,17 @@ export class Container {
...
+
     this.name = props.name ?? 'main';
     this.image = props.image;
     this.port = props.port;
     this._command = props.command;
     this._args = props.args;
-    this._readiness = props.readiness;
-    this._liveness = props.liveness;
+    this._readiness = props.readiness ?? ( props.port ? defaultProbeConfiguration : undefined);
+    this._liveness = props.liveness ?? ( props.port ? defaultProbeConfiguration : undefined);
```
iliapolo: @vinayak-kukreja After some more thought I think we shouldn't be configuring default liveness / readiness probes. 

I fear that in many cases, these default probes will not actually represent a healthy container, and by applying them, we are masking a would be validation violation that forces the application author to properly think about these probes. 

I do think though that we can use the TCP probe for the `startupProbe`, since a container is definitely not ready before its port is bounded. The `startupProbe` is safer because it doesn't play a role in the ongoing lifecycle of the container.

iliapolo (changes requested):

src/container.ts
vinayak-kukreja: I see, makes sense. Will update this.

iliapolo (approved):

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-23/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/1124)|

### Questions ?

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-22/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/1125)|

### Questions ?

PR #1126, opened 2022-09-11
feat: expose L1 constructs

### LINKED ISSUE #937: Publish core L1 resources as its own library
Currently users have to run `cdk8s import` to get access to the core kubernetes resource API. 

This is problematic:

- Its not intuitive, often missed by first time users.
- It requires setting up a re-import task as part of a build process, to get bug fixes and updates in the generated code.
- Its bulky, in some languages the amount of boiler plate code it adds is pretty big.

Instead, lets vend dedicated libraries that contain these L1's, which can be updated in the normal dependency upgrade process.

iliapolo:
Instead of vending them as separate libraries, lets just expose them from `cdk8s-plus`, as it already vends them.

Resolves https://github.com/cdk8s-team/cdk8s/issues/937

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,3 +23,4 @@ export * from './network-policy';
 export * from './namespace';
 
 export * from './api-resource.generated';
+export * as k8s from './imports/k8s';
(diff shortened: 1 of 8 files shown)


vinayak-kukreja (approved):
Nice 🚀

PR #1146, opened 2022-09-14
feat(serviceaccount, pod): set automountServiceAccountToken to false

### LINKED ISSUE #818: Configure default `automountServiceAccountToken` in pods and service accounts
Should be set to `false`.

### Additional resources

- https://hub.datree.io/built-in-rules/prevent-service-account-token-auto-mount

vinayak-kukreja:
Kubernetes provisions a service account when creating a Pod. This service account's secret token is mounted within the Pod at runtime. 
Many containerized applications do not require direct access to the service account. Hence, setting `automountServiceAccountToken` to `false` to align with best practices.

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/818


src/pod.ts
```diff
@@ -149,9 +149,9 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
    */
   public toSubjectConfiguration(): rb.SubjectConfiguration {
 
-    if (!this.serviceAccount && !this.automountServiceAccountToken) {
```
iliapolo: Why did you remove the condition? `automountServiceAccountToken` can still be `true` in case the user configured it explicitly

iliapolo (approved):

PR #1174, opened 2022-09-20
feat(horizontal-pod-autoscaler): Introduce `HorizontalPodAutoscaler` construct

### LINKED ISSUE #819: `HorizontalPodAutoscaler` L2
https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/

ryparker:
Resolves #819


src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,823 @@
...
+   * The workload to scale up or down.
+   *
+   * Scalable workload types:
+   * * Deployment
+   * * ReplicaSet
+   * * ReplicationController
+   * * StatefulSet
+   *
+   * Cannot be used with workloads that do not scale, such as DaemonSet.
+   */
```
iliapolo: It's also true for `Job` and `CronJob` I imagine, which are workloads as well. Accepting a `Workload` here opens the door for user error that we could prevent.
The way to address this is to instead of accepting a `Workload`, we accept a new `Scalable` interface:

```ts
export interface ScalingTarget {
  readonly kind: string;
  readonly apiVersion: string;
  readonly name: string;
  readonly containers: Container[];
}

export interface Scalable {
  toScalingTarget(): ScalingTarget;
}
```

And make the scalable resources that can be passed as targets implement this interface.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+   * at least one Object or External metric is configured.
+   *
+   * @default 1
+   */
+  readonly minReplicas?: number;
+  /**
+   * The metric conditions that trigger a scale up or scale down.
+   *
+   * @default
+   * No metrics are configured. Only scales on the requests or limits defined on the target.
```
iliapolo: Our convention is to use `default -` when we describe a behavior.

```suggestion
   * @default - No metrics are configured. Only scales on the resource requests or limits defined on the target.
```

Also, what is the operator of the conditions? OR? AND?

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+   */
+  readonly metrics?: Metric[];
+  /**
+   * The scaling behavior when scaling up.
+   *
+   * @default
+   * Is the higher of:
+   * * Increase no more than 4 pods per 60 seconds
+   * * Double the number of pods per 60 seconds
```
iliapolo: Not sure this default makes sense...interested why you made this decision, and what is the kubernetes default.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+   */
+  readonly scaleUp?: ScalingRules;
+  /**
+   * The scaling behavior when scaling down.
+   *
+   * @default
+   * Scale down to minReplica count with a 5 minute stabilization window.
```
iliapolo: Same as above

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+ * * StatefulSet
+ *
+ * :warning: HorizontalPodAutoscaler cannot be used with a DaemonSet.
+ *
+ * ### Targets that already have a replica count defined.
+ *
+ * It is recommended to remove any replica counts from the target resource before using a
+ * HorizontalPodAutoscaler. If this isn't done, then any time a change to that object is applied,
+ * Kubernetes will scale the current number of Pods to the value of the target.replicas key. This
+ * may not be desired and could lead to unexpected behavior.
```
iliapolo: This is a problem because we set default replicas counts, for example:

https://github.com/cdk8s-team/cdk8s-plus/blob/31b87be4d51619d9acc1fd0845a87682dc2539d6/src/deployment.ts#L159

So basically a user is currently unable to "remove" replica counts. It would be really nice if these defaults are only applied in case the workload is not a target of any HPA. You can achieve this with some fancy construct tree introspection. This will also serve as a validation that targets should not have replica counts on them. Technically it means deferring setting the default value to the `_toKube` methods, instead of the constructor.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+  /**
+   * The workload to scale up or down.
+   *
+   * Scalable workload types:
+   * * Deployment
+   * * ReplicaSet
+   * * ReplicationController
+   * * StatefulSet
+   *
+   * Cannot be used with workloads that do not scale, such as DaemonSet.
```
iliapolo: As output properties, their docs should be different:

```suggestion
   * The workload scaled by this HPA.
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+   * The maximum number of replicas that can be scaled up to.
+   */
+  public readonly maxReplicas: number;
+  /**
+   * The minimum number of replicas that can be scaled down to.
+   *
+   * Can be set to 0 if the alpha feature gate `HPAScaleToZero` is enabled and
+   * at least one Object or External metric is configured.
+   *
+   * @default 1
```
iliapolo: Again, output properties should speak to what is currently defined on the instance, not what are the possible configurations. 

```suggestion
   * The minimum number of replicas that can be scaled down to.
```

This repeats throughout the docs of output properties, please make sure to go through all of them.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+  constructor(scope: Construct, id: string, props: HorizontalPodAutoscalerProps) {
+    super(scope, id);
+    this.apiObject = new k8s.KubeHorizontalPodAutoscalerV2(this, 'Resource', {
+      metadata: props.metadata,
+      spec: Lazy.any({ produce: () => this._toKube() }),
+    });
+
+    if (props.target.kind === 'DaemonSet') {
+      throw new Error('HorizontalPodAutoscaler cannot be used with workloads that do not scale, such as a DaemonSet.');
+    }
```
iliapolo: Will not be necessary with the new `Scalable` interface.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+    const hasResourceConstraints = props.target.containers.some((c) => this._hasRequestsOrLimits(c));
+    if (!hasResourceConstraints && !props.metrics) {
+      throw new Error('Every container in the HorizontalPodAutoscaler target must have CPU or memory resources defined');
```
iliapolo: A bit misleading since this only holds in case `metrics` are not defined, mention this in the error message.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+    }
+    if (props?.minReplicas && props.minReplicas > props.maxReplicas) {
+      throw new Error("'minReplicas' must be less than or equal to maxReplicas in order for HorizontalPodAutoscaler to scale.");
```
iliapolo: Mention the violating values in the error message. I usually use parentheses:

 ```suggestion
      throw new Error(`'minReplicas' (${props.minReplicas}) must be less than or equal to 'maxReplicas' *(${props.maxReplicas}) in order for HorizontalPodAutoscaler to scale.);
```

File: src/horizontal-pod-autoscaler.ts [hunk 11]
```diff
@@ -0,0 +1,866 @@
...
+    }
+    if (props?.scaleUp?.stabilizationWindow && !this._isStabilizationWindowWithinRange(props.scaleUp.stabilizationWindow)) {
+      throw new Error("'scaleUp.stabilizationWindow' must be more than 0 seconds and no longer than 1 hour.");
```
iliapolo: Mention current value in error message

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+    }
+    if (props?.scaleDown?.stabilizationWindow && !this._isStabilizationWindowWithinRange(props.scaleDown.stabilizationWindow)) {
+      throw new Error("'scaleDown.stabilizationWindow' must be more than 0 seconds and no longer than 1 hour.");
```
iliapolo: And here as well

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+  public readonly maxReplicas: number;
+  /**
+   * The minimum number of replicas that can be scaled down to.
+   *
+   * Can be set to 0 if the alpha feature gate `HPAScaleToZero` is enabled and
+   * at least one Object or External metric is configured.
+   *
+   * @default 1
+   */
+  public readonly minReplicas?: number;
```
iliapolo: Since you apply the default in the constructor...(like you should)

```suggestion
  public readonly minReplicas: number;
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+        duration: Duration.minutes(1),
+      },
+    ],
+  };
+  private readonly _defaultScaleDown = {
+    strategy: ScalingStrategy.MAX_CHANGE,
+    stabilizationWindow: Duration.minutes(5),
+    policies: [
+      {
+        replicas: Replicas.absolute(this.minReplicas ?? 1),
```
iliapolo:
```suggestion
        replicas: Replicas.absolute(this.minReplicas),
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+      ...props?.scaleDown,
+    };
+  }
+
+  /**
+   * Validate a list of scaling policies.
+   * @internal
+   */
+  private _validateScalingPolicies(direction: ScalingDirection, policies: ScalingPolicy[]) {
+    return policies.map((p) => {
```
iliapolo: Why `return` on a validation function? Its confusing because it gives the impression the return value is being used somewhere, but isn't. 

```suggestion
    policies.forEach((p) => {
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+   * @internal
+   */
+  private _hasRequestsOrLimits(container: Container): Boolean {
+    const hasRequests = container.resources?.cpu?.request || container.resources?.memory?.request;
+    const hasLimits = container.resources?.cpu?.limit || container.resources?.memory?.limit;
+    return Boolean(hasRequests || hasLimits);
+  }
+
+  /**
+   * @internal
```
iliapolo, line 297: Would be good to add a k8s doc reference here mentioning this range.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+   */
+  private _isStabilizationWindowWithinRange(window?: Duration): Boolean {
+    const windowSeconds = window?.toSeconds();
+    const isWithinRange = windowSeconds && (0 < windowSeconds && windowSeconds <= 3600);
+    return Boolean(isWithinRange);
+  }
+
+  /**
+   * @internal
```
iliapolo, line 311: Same

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+  public _toKube(): k8s.HorizontalPodAutoscalerSpecV2 {
+    return {
+      maxReplicas: this.maxReplicas,
+      minReplicas: this.minReplicas,
+      scaleTargetRef: {
+        apiVersion: this.target.apiVersion,
+        kind: this.target.kind,
+        name: this.target.name,
+      },
+      metrics: this.metrics?.map(m => m?._toKube()),
```
iliapolo:
```suggestion
      metrics: this.metrics?.map(m => m._toKube()),
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+      behavior: {
+        scaleUp: {
+          policies: this.scaleUp?.policies?.map(p => ({
+            type: p.replicas.type,
+            value: p.replicas.value,
+            periodSeconds: p.duration?.toSeconds() ?? Duration.seconds(15).toSeconds(),
```
iliapolo: Why is the 15 second duration default also here? You already define it here:

https://github.com/cdk8s-team/cdk8s-plus/blob/df6603c9ea690a8a08b44768e5454945cbac0ddd/src/horizontal-pod-autoscaler.ts#L193-L195

Try avoiding duplicate definitions.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+  public readonly metrics?: Metric[];
+  /**
+   * The scaling behavior when scaling up.
+   *
+   * @default
+   * Is the higher of:
+   * * Increase no more than 4 pods per 60 seconds
+   * * Double the number of pods per 60 seconds
+   */
+  public scaleUp: ScalingRules;
```
iliapolo:
```suggestion
  public readonly scaleUp: ScalingRules;
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+  /**
+   * The scaling behavior when scaling down.
+   *
+   * @default
+   * Scale down to minReplica count with a 5 minute stabilization window.
+   */
+  public scaleDown: ScalingRules;
```
iliapolo:
```suggestion
  public readonly scaleDown: ScalingRules;
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+      metrics: this.metrics?.map(m => m?._toKube()),
+      behavior: {
+        scaleUp: {
+          policies: this.scaleUp?.policies?.map(p => ({
+            type: p.replicas.type,
+            value: p.replicas.value,
+            periodSeconds: p.duration?.toSeconds() ?? Duration.seconds(15).toSeconds(),
+          })),
+          selectPolicy: this.scaleUp?.strategy,
+          stabilizationWindowSeconds: this.scaleUp?.stabilizationWindow?.toSeconds(),
```
iliapolo:
```suggestion
          stabilizationWindowSeconds: this.scaleUp.stabilizationWindow?.toSeconds(),
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+        },
+        scaleDown: {
+          policies: this.scaleDown?.policies?.map(p => ({
+            type: p.replicas.type,
+            value: p.replicas.value,
+            periodSeconds: p.duration?.toSeconds() ?? Duration.seconds(15).toSeconds(),
+          })),
+          selectPolicy: this.scaleDown?.strategy,
+          stabilizationWindowSeconds: this.scaleDown?.stabilizationWindow?.toSeconds(),
```
iliapolo:
```suggestion
          stabilizationWindowSeconds: this.scaleDown.stabilizationWindow?.toSeconds(),
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+          selectPolicy: this.scaleUp?.strategy,
+          stabilizationWindowSeconds: this.scaleUp?.stabilizationWindow?.toSeconds(),
+        },
+        scaleDown: {
+          policies: this.scaleDown?.policies?.map(p => ({
+            type: p.replicas.type,
+            value: p.replicas.value,
+            periodSeconds: p.duration?.toSeconds() ?? Duration.seconds(15).toSeconds(),
+          })),
+          selectPolicy: this.scaleDown?.strategy,
```
iliapolo:
```suggestion
          selectPolicy: this.scaleDown.strategy,
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+      },
+    };
+  }
+}
+
+
+/**
+ * Metric source types.
+ */
+export enum MetricSourceType {
```
iliapolo: You don't need this enum, see comments below.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,866 @@
...
+  /**
+   * Metric that tracks the CPU of a container. This metric
+   * will be tracked across all pods of the current scale target. Only a single
+   * container may be tracked in each pod. In other words, a pod that hosts two
+   * of the same container, will be ignored.
+   *
+   */
+  public static containerCpu(options: MetricContainerResourceOptions): Metric {
+    return new Metric({
+      type: MetricSourceType.CONTAINER_RESOURCE,
```
iliapolo:
```suggestion
      type: 'ContainerResource',
```

And same for the other functions. Then you can get rid of the `MetricSourceType` enum, which in any case shouldn't have been exported.

test/horizontal-pod-autoscaler.errors.test.ts
```diff
@@ -0,0 +1,229 @@
+import { Testing, Duration } from 'cdk8s';
```
iliapolo: No need for a separate file

test/horizontal-pod-autoscaler.defaults.test.ts
```diff
@@ -0,0 +1,220 @@
+import { Testing, ApiObject, Duration } from 'cdk8s';
```
iliapolo: No need for a separate file

test/horizontal-pod-autoscaler.test.ts
```diff
@@ -0,0 +1,125 @@
...
+test('creates HPA with expected spec, when metrics, scaleUp, and scaleDown are all configured', () => {
+  const chart = Testing.chart();
+  const deployment = new kplus.Deployment(chart, 'Deployment', { containers: [{ image: 'pod' }] });
+
+  new kplus.HorizontalPodAutoscaler(chart, 'Hpa', {
+    target: deployment,
+    maxReplicas: 10,
+    minReplicas: 2,
+    metrics: [
+      kplus.Metric.resourceCpu(kplus.MetricTarget.averageUtilization(50)),
```
iliapolo, line 734: For consistency: 

```suggestion
      kplus.Metric.resourceCpu({
        target: kplus.MetricTarget.averageUtilization(50),
      }),
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+}
+
+/**
+ * Options for `Metric.object()`
+ */
+export interface MetricObjectOptions extends MetricOptions {
+  /**
+   * Resource where the metric can be found.
+   */
+  readonly object: Resource;
```
iliapolo: Always try accepting interfaces instead of concrete classes. Makes it more flexible. 

```suggestion
  readonly object: IResource;
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+          name: options.name,
+          selector: options?.labelSelector?._toKube(),
+        },
+        target: options.target._toKube(),
+      },
+    });
+  }
+
+  /**
+  * Metric that describes a kubernetes object
```
iliapolo: Not entirely accurate, at least not based on the API.

```suggestion
  * Metric that describes a metric of a kubernetes object
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+          name: options.name,
+          selector: options?.labelSelector?._toKube(),
+        },
+        target: options.target._toKube(),
+      },
+    });
+  }
+
+  /**
+   * Tracks the available memory of a target.
```
iliapolo:
```suggestion
   * Tracks the available CPU of the pods in a target.
```

I would also add this note from kubernetes docs:

> Note: Since the resource usages of all the containers are summed up the total pod utilization may not accurately represent the individual container resource usage. This could lead to situations where a single container might be running with high usage and the HPA will not scale out because the overall pod usage is still within acceptable limits.

And the same of the other functions.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+      type: MetricSourceType.RESOURCE,
+      resource: {
+        name: ResourceType.CPU,
+        target: target._toKube(),
+      },
+    });
+  }
+
+  /**
+   * Tracks the available CPU of a target.
```
iliapolo:
```suggestion
   * Tracks the available memory of a target.
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+   * Local ephemeral storage
+   */
+  EPHEMERAL_STORAGE = 'ephemeral-storage',
+}
+
+
+/**
+ * How to interpret the metric value.
+ */
+export enum MetricTargetType {
```
iliapolo: Same as `MetricSourceType`, you don't need this.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+ * MetricTarget.averageUtilization(70); // 70% average utilization
+ *
+ */
+export class MetricTarget {
+  /**
+  * Target a specific target value.
+  *
+  * @param value The target value. Floats can be passed within a string. i.g. `"1.2"`
+  */
+  public static value(value: string): MetricTarget {
```
iliapolo: You can pass floats to a `number` type...

```suggestion
  public static value(value: number): MetricTarget {
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+    return new MetricTarget({
+      type: MetricTargetType.AVERAGE_VALUE,
+      averageValue: k8s.Quantity.fromString(averageValue),
+    });
+  }
+
+  /**
+   * Target a percentage value across all relevant pods.
+   *
+   * @param averageUtilization The percentage of the utilization metric. i.g. `50` for 50%.
```
iliapolo:
```suggestion
   * @param averageUtilization The percentage of the utilization metric. e.g. `50` for 50%.
```

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+  /**
+   * @internal
+   */
+  public _toKube(): k8s.MetricSpecV2 {
+    return this.metric;
+  }
+
+}
+
+export enum ResourceType {
```
iliapolo: Not needed

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+  public _toKube(): k8s.MetricSpecV2 {
+    return this.metric;
+  }
+}
+
+
+/**
+ * The direction to scale in.
+ */
+export enum ScalingDirection {
```
iliapolo: Not needed, or at least, no need to export

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+  /**
+   * The target metric value that will trigger scaling.
+   */
+  readonly target: MetricTarget;
+  /**
+  * The name of the metric to scale on.
+  */
+  readonly name: string;
+  /**
+   * A selector to find a metric by labels.
```
iliapolo: What is the default? can this cause conflicts in case the metrics selected by the selector don't match the given name?

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+/**
+ * A metric condition that HorizontalPodAutoscaler's scale on.
+ */
+export class Metric {
+
+  /**
+   * Metric that tracks the CPU of a container. This metric
+   * will be tracked across all pods of the current scale target. Only a single
+   * container may be tracked in each pod. In other words, a pod that hosts two
+   * of the same container, will be ignored.
```
iliapolo: > In other words, a pod that hosts two of the same container, will be ignored.

What does this mean? how can a pod have two of the same container? same name?

Also, does this track a container across all pods in the cluster? or across all replicas of the same pod?

In general, I found this a little hard to reason about, and what eventually helped me is kubernetes docs:

> The HorizontalPodAutoscaler API also supports a container metric source where the HPA can track the resource usage of individual containers across a set of Pods, in order to scale the target resource. This lets you configure scaling thresholds for the containers that matter most in a particular Pod. For example, if you have a web application and a logging sidecar, you can scale based on the resource use of the web application, ignoring the sidecar container and its resource use.

See if you can incorporate this paragraph in the docstring somehow, it helps a lot. Don't worry about bloated doc strings, or duplicated text, in this case its a good thing.

src/horizontal-pod-autoscaler.ts
iliapolo: Also, tracks CPU across what exactly? All containers in the pod?

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+    }
+    if (props?.scaleUp?.stabilizationWindow && !this._isStabilizationWindowWithinRange(props.scaleUp.stabilizationWindow)) {
+      throw new Error("'scaleUp.stabilizationWindow' must be more than 0 seconds and no longer than 1 hour.");
+    }
+    if (props?.scaleDown?.stabilizationWindow && !this._isStabilizationWindowWithinRange(props.scaleDown.stabilizationWindow)) {
+      throw new Error("'scaleDown.stabilizationWindow' must be more than 0 seconds and no longer than 1 hour.");
+    }
+    if (props?.scaleUp?.policies?.length) {
+      this._validateScalingPolicies(ScalingDirection.UP, props.scaleUp.policies);
+      props.scaleUp.policies = this._setDefaultsOnScalingPolicies(props.scaleUp.policies);
```
iliapolo: It's not good practice to mutate properties. If we can do it, the user can also mistakenly do it. All structs should only contain `readonly` properties. Find another way to set these defaults...

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+  /**
+   * Scale down
+   */
+  DOWN = 'scaleDown',
+}
+
+/**
+ * Defines the scaling behavior for one direction.
+ */
+export interface IScalingRules {
```
iliapolo: Why not `ScalingRules`? I suspect because jsii doesn't allow non `readonly` properties on structs? It enforces that for a good reason, don't workaround it :)

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+   */
+  readonly duration?: Duration;
+}
+
+
+/**
+ * Describes how much change will be applied when a ScalingPolicy is selected
+ * for scaling.
+ */
+export interface ReplicasToScale {
```
iliapolo: You shouldn't need this interface, wherever you use it, you should be using `Replicas` as is instead.

src/horizontal-pod-autoscaler.ts
```diff
@@ -0,0 +1,865 @@
...
+   *
+   * @deprecated - Omit the ScalingRule instead
+   */
+  DISABLED = 'Disabled',
+}
+
+/**
+* The type of scaling policy that could be used to make scaling decisions.
+*/
+export enum ScalingPolicyType {
```
iliapolo: Again not needed.

iliapolo (changes requested):
Looks good. We also need user facing docs in the website though, similar to [Network Policy](https://cdk8s.io/docs/latest/plus/network-policy/) for example.

src/horizontal-pod-autoscaler.ts
ryparker: This is actually the Kubernetes default. All the default values I have so far are only from Kubernetes.

src/horizontal-pod-autoscaler.ts
ryparker: This is a Kubernetes default.

src/horizontal-pod-autoscaler.ts
ryparker: Totally agree. I'll see what I can do.

src/horizontal-pod-autoscaler.ts
ryparker: This is a misunderstanding on my part. I'll lean more on the Kubernetes description here.

src/horizontal-pod-autoscaler.ts
ryparker: Yeah It felt gross doing it. I was having some struggles with making JSII and TS happy. Was also trying to avoid setting the default values in the `_toKube()`. I'll rework it.

src/horizontal-pod-autoscaler.ts
ryparker: Got it 👍🏻

ryparker:
> Looks good. We also need user facing docs in the website though, similar to [Network Policy](https://cdk8s.io/docs/latest/plus/network-policy/) for example.

website overview page here: https://github.com/cdk8s-team/cdk8s/pull/1001

src/horizontal-pod-autoscaler.ts
ryparker: Got this wired up however I wasn't able to get around the JSII error stating that "`Scalable` contains a behavior." So I was forced to make it an interface: `IScalable`. Any pointers on how I can avoid this?

src/horizontal-pod-autoscaler.ts
ryparker: Looks like it only uses the selector for specificity. If not provided it will only use the provided metric name.

https://github.com/kubernetes/kubernetes/blob/00532e305a3958ec08ce4453d9f234c4a0fbbbf8/staging/src/k8s.io/api/autoscaling/v1/types_swagger_doc_generated.go#L217

ryparker (commented):
Thanks @iliapolo. Updated with the suggestions.

ryparker:
Moved HPA overview website doc from cdk8s repo to this PR and changed target branch to the new v25 branch

docs/plus/horizontal-pod-autoscaler.md
```diff
@@ -0,0 +1,192 @@
+# HorizontalPodAutoscaler
+
+HorizontalPodAutoscaler allows your services to scale up when demand is high and scale down when they are no longer needed.
+
+!!! tip ""
+    [API Reference](../reference/cdk8s-plus-24/typescript.md#horizontalpodautoscaler)
```
iliapolo:
```suggestion
    [API Reference](../reference/cdk8s-plus-25/typescript.md#horizontalpodautoscaler)
```

docs/plus/horizontal-pod-autoscaler.md
```diff
@@ -0,0 +1,192 @@
...
+
+const hpa = new kplus.HorizontalPodAutoscaler(chart, 'BookstoreWebsiteHpa', {
+  target: bookstoreWebsite,
+  maxReplicas: 10,
+  metrics: [
+     kplus.Metric.resourceCpu(kplus.MetricTarget.averageUtilization(70)),
+  ],
+ });
+```
+
```
iliapolo, line 85: We should mention the note [k8s docs](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-resource-metrics) provides for resource metrics:

> Note: Since the resource usages of all the containers are summed up the total pod utilization may not accurately represent the individual container resource usage. This could lead to situations where a single container might be running with high usage and the HPA will not scale out because the overall pod usage is still within acceptable limits.

docs/plus/horizontal-pod-autoscaler.md
```diff
@@ -0,0 +1,192 @@
...
+### Pods Metrics
```
iliapolo, line 86: What about [Container Resource Metrics](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#container-resource-metrics)? I think it should go right after the resource metrics.

docs/plus/horizontal-pod-autoscaler.md
```diff
@@ -0,0 +1,192 @@
...
+
+```typescript
+import * as kplus from 'cdk8s-plus-25';
+
+
+const hpa = new kplus.HorizontalPodAutoscaler(chart, 'BookstoreWebsiteHpa', {
+  target: bookstoreWebsite,
+  maxReplicas: 10,
+  metrics: [
+     kplus.Metric.pods(kplus.MetricTarget.averageValue(70)),
```
iliapolo: This seems outdated with the API? Also if you mention a specific example (transactions-processed-per-second) make sure to use it in the code, I was looking for it.

docs/plus/horizontal-pod-autoscaler.md
```diff
@@ -0,0 +1,192 @@
...
+
+```typescript
+import * as kplus from 'cdk8s-plus-25';
+
+
+const hpa = new kplus.HorizontalPodAutoscaler(chart, 'BookstoreWebsiteHpa', {
+  target: bookstoreWebsite,
+  maxReplicas: 10,
+  metrics: [
+     kplus.Metric.object(kplus.MetricTarget.averageValue(70)),
```
iliapolo: Same

docs/plus/horizontal-pod-autoscaler.md
```diff
@@ -0,0 +1,192 @@
...
+
+```typescript
+import * as kplus from 'cdk8s-plus-25';
+
+
+const hpa = new kplus.HorizontalPodAutoscaler(chart, 'BookstoreWebsiteHpa', {
+  target: bookstoreWebsite,
+  maxReplicas: 10,
+  metrics: [
+     kplus.Metric.external(kplus.MetricTarget.averageValue(70)),
```
iliapolo: Same

src/deployment.ts
```diff
@@ -215,7 +216,7 @@ export class Deployment extends workload.Workload {
    */
   public _toKube(): k8s.DeploymentSpec {
     return {
-      replicas: this.replicas,
+      replicas: this.replicas ?? 2,
```
iliapolo: Not sure I follow this, won't this interfere with HPA if configured on this deployment? We spoke about providing a default replica count **only if** an HPA isn't defined, but this doesn't seem to consider it.

src/deployment.ts
```diff
@@ -227,6 +228,18 @@ export class Deployment extends workload.Workload {
...
+  /**
+   * @see IScalable.toScalingTarget()
+   */
+  public toScalingTarget(): ScalingTarget {
+    return {
+      kind: this.apiObject.kind,
+      apiVersion: this.apiObject.apiVersion,
+      name: this.name,
+      containers: this.containers,
+      replicas: this.replicas ?? 0,
```
iliapolo: Why use `0` as a default? Seems dangerous, won't `undefined` work just as well?

iliapolo (changes requested):

src/deployment.ts
ryparker: Yeah something needs to change here. When writing this I was under the impression that `_toKube()` would only be called if it wasn't associated with an HPA since we now have `toScalingTarget()`. I didn't consider that `_toKube()` can still be called by the chart/other associated constructs. I'll have to think about this.

src/deployment.ts
ryparker: Sure we can accept undefined here. I only did this to keep the interface type tight. Although out of curiosity can you explain why setting `0` here would be dangerous? To prevent unexpected `??` operator behavior downstream?

UPDATE: change has been applied

ryparker: Updated the `ScalingTarget` to accept undefined

iliapolo (approved):

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-23/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/1300)|

### Questions ?

PR #1176, opened 2022-09-20
feat(workload): `spread` property for easy `hostname` and `zone` spreading

### LINKED ISSUE #890: Provide a `spread` property on workloads to allow for easy anit-affinity
### Discussed in https://github.com/cdk8s-team/cdk8s-plus/discussions/822

<div type='discussions-op-text'>

<sup>Originally posted by **iliapolo** June  9, 2022</sup>
It's good practice to enforce that multiple pods of the same workload don't end up on the same host during scheduling. 

Should `cdk8s-plus` configure this by default? 

See https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#more-practical-use-cases

### Additional questions

- Does kubernetes do this by default? try to?</div>

vinayak-kukreja:
Following Kubernetes [guidance](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#internal-default-constraints) to spread pods across nodes and zones for a workload. Adding a default spread property to workloads. 

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/890

### DIFF
--- a/src/workload.ts
+++ b/src/workload.ts
@@ -26,8 +26,8 @@ export interface WorkloadProps extends pod.AbstractPodProps {
   /**
    * Automatically spread pods across hostname and zones.
    *
-   * @default true
    * @see https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#internal-default-constraints
+   * @default true
    */
   readonly defaultSpread?: boolean;
 }
@@ -141,6 +141,9 @@ export abstract class Workload extends pod.AbstractPod {
       {
         this.scheduling.spread({
           topology: pod.Topology.HOSTNAME,
+        });
+        this.scheduling.spread({
+          topology: pod.Topology.ZONE,
         });
       }
     };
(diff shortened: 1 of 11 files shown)


ryparker (commented):
👍🏻

src/workload.ts
```diff
@@ -23,6 +23,13 @@ export interface WorkloadProps extends pod.AbstractPodProps {
    */
   readonly select?: boolean;
 
+  /**
+   * Automatically spread pods across hostname and zones.
+   *
+   * @see https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#internal-default-constraints
+   * @default true
+   */
+  readonly defaultSpread?: boolean;
```
iliapolo: Lets call this just `spread`, similar to `select`.

src/workload.ts
```diff
@@ -23,6 +23,13 @@ export interface WorkloadProps extends pod.AbstractPodProps {
    */
   readonly select?: boolean;
 
+  /**
+   * Automatically spread pods across hostname and zones.
+   *
+   * @see https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#internal-default-constraints
+   * @default true
```
iliapolo: The default should be `false` to be more developer friendly, i.e allowing default constructs to run on a dev cluster with one node.

iliapolo (changes requested):

iliapolo (approved):

PR #1177, opened 2022-09-20
feat(pod): isolate pod of all network connections

### LINKED ISSUE #891: Provide a property for pods to be network isolated
### Discussed in https://github.com/cdk8s-team/cdk8s-plus/discussions/821

<div type='discussions-op-text'>

<sup>Originally posted by **iliapolo** June  9, 2022</sup>
There's a way to create network policies that isolate pods for egress / ingress. That is, pods won't be able to accept or initiate network traffic.

- See https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-and-all-egress-traffic

Intuitively, as far as security and architectural best practices, this feels like it should be the default. Any network connection should be explicitly defined and allowed.

`cdk8s-plus` allows us to apply this default very easily, question is - should we? </div>

vinayak-kukreja:
Kubernetes has [functionality](https://kubernetes.io/docs/concepts/services-networking/network-policies/#default-deny-all-ingress-and-all-egress-traffic) to set the default behavior of network policy to disable all egress and ingress connections of pods. Updating our default behavior in this PR. 

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/891


vinayak-kukreja:
**Question**

I was initially trying to do this with the help of [Namespaces](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-24/main/src/namespace.ts#L45). But I see we do not create this resource explicitly in our workflow and just use the [string](https://github.com/cdk8s-team/cdk8s-core/blob/2.x/src/metadata.ts#L53) instead. Having the default network policy initialized when a namespace was created would help to have this policy work for each pod within that namespace. 

The problem with doing this currently is,
* If we have multiple resources(like deployment, job, etc) using the same name for the namespace. Then if we add logic to create a namespace in each of these resources, it would duplicate namespace in the manifest, due to multiple instances of it. I am not sure if there is a single place where the other resources can keep track of an already created namespace.
* I did not find any place where we can create a default network policy for `default` namespace. 

The solution I could think of was to create a Namespace instance and pass it to the resources instead of the string in metadata. Let me know what you think?

ryparker (commented):
Looks good 👍🏻

src/pod.ts
```diff
@@ -1753,4 +1747,21 @@ export class PodConnections {
...
+      // the policy must be defined in the namespace of the pod
+      // so it can select it.
+      metadata: Lazy.any({ produce: () => {return { namespace: this.instance.metadata.namespace };} }),
+      egress: {
+        default: networkpolicy.NetworkPolicyTrafficDefault.DENY,
+      },
+      ingress: {
+        default: networkpolicy.NetworkPolicyTrafficDefault.DENY,
+      },
+    });
```
iliapolo: So for every pod in the same namespace, this will create the same network policy..a bit awkward.
Also, this wasn't the intent. I can see how this got mixed up because the issue didn't describe it well enough. 

The idea was to allow the user to easily create an isolated **pod**, not an entire **namespace**.
I do think namespace isolation is a cool feature as well, but this should be implemented in the [Namespace](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-24/main/src/namespace.ts) construct, not the pod.

src/pod.ts
```diff
@@ -1616,15 +1616,13 @@ export interface PodConnectionsAllowFromOptions {
  */
 export class PodConnections {
 
-  constructor(protected readonly instance: AbstractPod) {}
+  constructor(protected readonly instance: AbstractPod) {
+    this.defaultNetworkPolicy();
```
iliapolo: We shouldn't isolate by default, only upon user request.

iliapolo (changes requested):

src/pod.ts
vinayak-kukreja: I understand, will make the update.

src/pod.ts
vinayak-kukreja: Will update this.

ryparker:
FYI there's now a v25 branch, I'd consider changing the target to v25 and adding the `backport-to-k8s-24/main` label.

iliapolo (approved):

PR #1223, opened 2022-10-04
feat(stateful-set): `service` is now optional

iliapolo:
Based on my readings, a stateful set is usually accompanied by a headless service.

> See https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

This means we can provide a sane default here, and create a service on the user's behalf. The auto created service can then be accessed via the `.service` property, which is now made public.

### Misc

- Added `.port` property on a service to get easy access to the first port of the service, which is also many times the only one.


src/stateful-set.ts
```diff
@@ -127,16 +130,44 @@ export class StatefulSet extends workload.Workload {
...
     this.replicas = props.replicas ?? 1;
     this.strategy = props.strategy ?? StatefulSetUpdateStrategy.rollingUpdate(),
     this.podManagementPolicy = props.podManagementPolicy ?? PodManagementPolicy.ORDERED_READY;
     this.minReady = props.minReady ?? Duration.seconds(0);
 
-    this._service.select(this);
+    this.service.select(this);
+  }
+
+  private createHeadlessService() {
```
ryparker: ⛏️ Not sure what our rule is on appending `_` and if it goes on all private methods/props

```suggestion
  private _createHeadlessService() {
```

ryparker (changes requested):
Looks good. Small nitpick for you to consider.

src/stateful-set.ts
```diff
@@ -127,16 +130,44 @@ export class StatefulSet extends workload.Workload {
       metadata: props.metadata,
       spec: Lazy.any({ produce: () => this._toKube() }),
     });
-    this._service = props.service;
+    this.service = props.service ?? this.createHeadlessService();
```
ryparker: If you accept the other suggestion, then here's a follow up:

```suggestion
    this.service = props.service ?? this._createHeadlessService();
```

ryparker:
FYI there's now a v25 branch, I'd consider changing the target to v25 and adding the `backport-to-k8s-24/main` label.

ryparker (approved):

PR #1403, opened 2022-11-13
feat(volume): iSCSI & NFS volume types

dimakuz:


src/volume.ts
```diff
@@ -238,7 +264,7 @@ export class Volume extends Construct implements IStorage {
   };
 
 
-  private constructor(scope: Construct, id: string,
+  protected constructor(scope: Construct, id: string,
```
iliapolo: How come?

src/volume.ts
```diff
@@ -656,4 +682,98 @@ export enum HostPathVolumeType {
...
+   * readOnly here will force the ReadOnly setting in VolumeMounts. Defaults to false.
+   *
+   * @default false.
+   */
+  readonly readOnly?: boolean;
+
+  /**
+   * secretRef is the CHAP Secret for iSCSI target and initiator authentication
+   */
+  readonly secretRef?: IscsiVolumeSourceSecretRef;
```
iliapolo: Secrets are referenced different with cdk8s-plus, by using the [`Secret.fromSecretName`](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-25/main/src/secret.ts#L103) method.

```suggestion
  readonly secretRef?: ISecret;
```

This also allows managed secrets to be passed here (i.e ones that were created using `new Secret(...)`)

src/volume.ts
```diff
@@ -656,4 +682,98 @@ export enum HostPathVolumeType {
...
+ export interface NfsVolumeOptions {
+  /**
+   * path that is exported by the NFS server. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
+   */
+  readonly path: string;
+
+  /**
+   * readOnly here will force the NFS export to be mounted with read-only permissions. Defaults to false. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
+   *
+   * @default false. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
```
iliapolo:
```suggestion
   * @see https://kubernetes.io/docs/concepts/storage/volumes#nfs
   * @default false
```

We use the `@default` annotation for automatic API reference generation, so nothing should be appended to it that doesn't relate to default behavior.

src/volume.ts
```diff
@@ -656,4 +682,98 @@ export enum HostPathVolumeType {
...
-}
\ No newline at end of file
+}
+
+/**
+ * Options for an Nfs-based volume.
+ */
+ export interface NfsVolumeOptions {
+  /**
+   * path that is exported by the NFS server. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
```
iliapolo:
```suggestion
   * Path that is exported by the NFS server.
   * 
   * @see https://kubernetes.io/docs/concepts/storage/volumes#nfs
```

src/volume.ts
```diff
@@ -656,4 +682,98 @@ export enum HostPathVolumeType {
...
+   */
+  readonly path: string;
+
+  /**
+   * readOnly here will force the NFS export to be mounted with read-only permissions. Defaults to false. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
```
iliapolo:
```suggestion
   * Force the NFS export to be mounted with read-only permissions.
```

src/volume.ts
```diff
@@ -656,4 +682,98 @@ export enum HostPathVolumeType {
...
+   *
+   * @default false. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
+   */
+  readonly readOnly?: boolean;
+
+  /**
+   * server is the hostname or IP address of the NFS server. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
```
iliapolo:
```suggestion
   * The hostname or IP address of the NFS server. 
   *
   * @see https://kubernetes.io/docs/concepts/storage/volumes#nfs
```

src/volume.ts
```diff
@@ -656,4 +682,98 @@ export enum HostPathVolumeType {
...
+
+  /**
+   * iqn is the target iSCSI Qualified Name.
+   */
+  readonly iqn: string;
+
+  /**
+   * iscsiInterface is the interface Name that uses an iSCSI transport. Defaults to 'default' (tcp).
+   *
+   * @default 'default' (tcp).
```
iliapolo: Normally we surface default values to the manifest, so instead of just leaving this `undefined`, actually set the value to `tcp`. This way those reading the manifest don't have to wonder what the behavior is.

src/volume.ts
```diff
@@ -221,6 +221,32 @@ export class Volume extends Construct implements IStorage {
...
+   * Used to mount an NFS export into a Pod.
+   *
+   * @see https://kubernetes.io/docs/concepts/storage/volumes/#nfs
+   */
+  public static fromNfs(scope: Construct, id: string, name: string, options: NfsVolumeOptions): Volume {
+    return new Volume(scope, id, name, {
+      nfs: {
+        server: options.server,
+        path: options.path,
+        readOnly: options.readOnly ?? false,
```
iliapolo:
```suggestion
        ...options,
        readOnly: options.readOnly ?? false,
```

iliapolo (changes requested):

iliapolo:
@dimakuz Thanks for this! Please also add some unit tests though, and a small description in the PR body.

src/volume.ts
dimakuz: It's either that, or a static constructor that accepts `k8s.Volume` (which is basically making this constructor public.

One of the more frustrating things I found about this is that the class is completely locked down to extension while not offering all option that k8s API currently has.

dimakuz:
Thanks for the review! Will revise the PR soon.

src/volume.ts
dimakuz: Dropped in favor of changing IStorage

src/volume.ts
```diff
@@ -656,4 +686,99 @@ export enum HostPathVolumeType {
...
+   * Target iSCSI Qualified Name.
+   */
+  readonly iqn: string;
+
+  /**
+   * Interface Name that uses an iSCSI transport. Defaults to 'default' (tcp).
+   *
+   * @default Protocol.TCP
+   */
+  readonly iscsiInterface?: string;
```
iliapolo, line 755: Can this be typed as an enum?

iliapolo (commented):
@dimakuz Sorry the (very) long delay on this...

Once the merge conflicts are resolved I think we can merge this.
Are you still able to pick it up?

PR #1437, opened 2022-11-23
fix(container): default uid and gid might cause some images to break

### LINKED ISSUE #1339: Omit defaults for SecurityContext
According to k8s [docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#securitycontext-v1-core) the `runAsUser` field is taken from the image metadata if omitted. 

cdk8splus24(golang) however does set the security context for pod and container with default values: 
```bash
[...]
securityContext:
  allowPrivilegeEscalation: false
  privileged: false
  readOnlyRootFilesystem: true
  runAsGroup: 26000
  runAsNonRoot: true
  runAsUser: 25000
[...]
```

without explicitly specifying a security context using code like this:
```golang
dep.AddContainer(&cdk8splus24.ContainerProps{
		Name:  jsii.String(name),
		Image: jsii.String(*image),
		Ports: &containerports,
	})
```
Although I like following best practices in regards of security by default setting good value of >1000 for `runAsUser`, this breaks images.
While we can set for instance `readOnlyRootFileystem: false`, we can set `runAsUser` only to an int value. Omitting it results in cdk8s+ setting the default of 25000, so no way to have kubernetes pick the UID from a docker image with cdk8s generated manifests.

Is there any way to avoid having `runAsUser` to be added to the manifests at all?

iliapolo:
While these defaults make sense from a security perspective, they do appear to be breaking some images that rely on the image being started with the default image user.

We need to come up with a mechanism that can toggle L2 defaults on and off, but for now, lets just revert this. 
 
Fixes https://github.com/cdk8s-team/cdk8s-plus/issues/1339


ryparker (approved):

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-24/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/1442)|


### Questions ?

iliapolo:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|k8s-23/main|[](https://github.com/cdk8s-team/cdk8s-plus/pull/1443)|


### Questions ?

PR #1449, opened 2022-11-24
fix(volume): names can become too long

### LINKED ISSUE #1130: Volume names can be too long and invalid
Names can be up to 63 characters, so when names are generated from the node path, they are trimmed. Otherwise the manifest would be invalid.

When you create a volume from a config map, and the name is not set, the default will be `configmap-${configMap.name}`:

https://github.com/cdk8s-team/cdk8s-plus/blob/788128a6d58901c3b4bb42dc0ef62f396628d019/src/volume.ts#L131

But the name of the config map can be the generated name. So now the length can be up to 10+63=73 (length of 'configmap-' + 63) because the length of the name is not checked again. That actually happens quite easily if you have nested constructs and don't set names. The same can happen for other volume functions, `fromAwsElasticBlockStore`, `fromAzureDisk`, etc. 

Maybe a better solution would be using either the ID of the volume, or the ID of the config map instead of the config map name? Or maybe not require a name for the Volume, remove the defaults and generate it from the node path like other names?

iliapolo:
For some volume types we default its name by prepending the volume type to the name of the volume source (configmap, pvc...). Problem is that this can surpass the 63 character limit kubernetes imposes on names.

This PR trims the volume name if necessary, by slicing characters from the end. 

Fixes https://github.com/cdk8s-team/cdk8s-plus/issues/1130


src/volume.ts
```diff
@@ -242,6 +242,7 @@ export class Volume extends Construct implements IStorage {
     public readonly name: string,
     private readonly config: Omit<k8s.Volume, 'name'>) {
     super(scope, id);
+    this.name = name.length <= 63 ? name : name.slice(0, 63);
```
dimakuz: Can be even sliced here unconditionally

ryparker:
```suggestion
    this.name = name.slice(0, 63);
```

ryparker (approved):
Small suggestion but G2G either way.

dancmeyers:
`Names.toDnsLabel` drops from the _beginning_, because you're more likely to retain a unique value that way. And often the name to be sliced here will be generated from `ApiObject.name`, which is calculated in the same sort of way I believe, so the same ideas apply?

It's a bit awkward to do that with the whole name in the `constructor` because we'll chop off some or all of the bit that the function in question specifically added (`configmap-` etc).

It wouldn't _look_ as nice in code, but it feels like the "less likely to cause a failure" method would be to pull this logic out of the constructor into each of the calling functions that adds a prefix. They then know the length of their prefix, so they can explicitly slice enough characters off the beginning of the rest of the name that the total is kept to 63 characters or less.

iliapolo:
@dancmeyers The problem with putting it in each of the individual functions is also that it makes it easier to forget, this way is more robust. 

As for trimming from the beginning instead, the thing is that the beginning of the string is the uniqueness added by the specific volume, otherwise we might end up with the same name as the api object.

PR #1512, opened 2022-12-14
feat(pod): allow adding an existing `Container` object

### LINKED ISSUE #1508: Deployment doesn't accept Container objects
Hi
I'm following this [doc](https://pypi.org/project/cdk8s-plus/) and trying to use `Container.mount()`

The problem is that the `Deployment.containters` expects `ContainerProps`, but `ContainerProps` doesn't have `mount` method.

The only way to use the mount method is by getting the container object from `Deployment.containers` after the deployment is created.

So the given example doesn't work
```python
container = kplus.Container(
    image="node:14.4.0-alpine3.12",
    command=["node", "index.js", f"{port}"],
    port=port,
)

container.mount(app_path, app_volume)
deployment = kplus.Deployment(chart, "Deployment",
    replicas=3,
    containers=[container]
)

``` 
Results with 
```
TypeError: type of argument containers must be one of (Sequence[Union[cdk8s_plus_25.ContainerProps, Dict[str, Any]]], NoneType); got list instead
```

The way to do this is
```python
container_p = kplus.ContainerProps(
    image="node:14.4.0-alpine3.12",
    command=["node", "index.js", f"{port}"],
    port=port,
)

deployment = kplus.Deployment(chart, "Deployment",
    replicas=3,
    containers=[container_p]
)

container = deployment.containers[0]
container.mount(app_path, app_volume)
```

This would be easier if Deployment would accept Container, or if Container had a `.props` property to convert it

yfried:
Fixes: #1508

### DIFF
--- a/src/pod.ts
+++ b/src/pod.ts
@@ -105,10 +105,14 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
     return this;
   }
 
-  public addContainer(cont: container.ContainerProps | container.Container): container.Container {
-    const impl = (cont instanceof container.Container) ? cont : new container.Container(cont);
-    this._containers.push(impl);
+  public addContainer(cont: container.ContainerProps): container.Container {
+    const impl = new container.Container(cont);
+    this.attachContainer(impl)
     return impl;
+  }
+
+  public attachContainer(cont: container.Container) {
+    this._containers.push(cont);
   }
 
   public addInitContainer(cont: container.ContainerProps): container.Container {
@@ -317,7 +321,7 @@ export interface AbstractPodProps extends base.ResourceProps {
    *
    * @default - No containers. Note that a pod spec must include at least one container.
    */
-  readonly containers?: container.ContainerProps[] | container.Container[];
+  readonly containers?: container.ContainerProps[];
 
   /**
    * List of initialization containers belonging to the pod.
--- a/src/pod.ts
+++ b/src/pod.ts
@@ -139,12 +139,6 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
     return impl;
   }
 
-  public attachInitContainer(cont: container.Container) {
-    this._initContainers.push(cont);
-
-  }
-
-
   public addHostAlias(hostAlias: HostAlias): void {
     this._hostAliases.push(hostAlias);
   }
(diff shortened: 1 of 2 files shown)


iliapolo (changes requested):
@yfried Unfortunately we can't do this because union types don't translate well to other languages. In Java for example this would translate to accepting an `Object`, which is not a good experience. What we usually do is just add another method with a different name that accepts the different argument.

yfried:
> @yfried Unfortunately we can't do this because union types don't translate well to other languages. In Java for example this would translate to accepting an `Object`, which is not a good experience. What we usually do is just add another method with a different name that accepts the different argument.

So instead you meant to have `attachConainer(cont: Container)` and `addContainer(cont: ContainerProps)`?
It's easy to do technically but I'm concerned that these 2 methods will look the same to new users (like me), and would raise a whole new bunch of questions/frustrations

iliapolo:
@yfried I agree adding another method might be confusing, but we can clarify the confusion with documentation. 

Ideally we would have the `addContainer` method accept a `Container`, and then have another method called `createContainer` that would accept `ContainerProps` and create the `Container`. But unfortunately we cannot break the API.

So given our limitations, I think adding the `attachXXX` methods will be an OK solution.

iliapolo (changes requested):
@yfried Lets also add `attachInitContainer`, and we need some unit tests added here: https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-25/main/test/pod.test.ts#L138

iliapolo (approved):

iliapolo:
@yfried looks good. Just need the DCO: https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-25/main/CONTRIBUTING.md#developer-certificate-of-origin-dco

src/pod.ts
```diff
@@ -107,10 +107,14 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
 
   public addContainer(cont: container.ContainerProps): container.Container {
     const impl = new container.Container(cont);
-    this._containers.push(impl);
+    this.attachContainer(impl)
```
iliapolo:
```suggestion
    this.attachContainer(impl);
```

iliapolo:
@yfried The build is failing because it detected a diff between your code and the one our linters expect. Can you run `yarn build` locally to auto-fix it?

PR #1519, opened 2022-12-15
feat(role): imported resources can be granted permissions on

### LINKED ISSUE #1516: Imported resources cannot be granted permissions on
It should be possible to do the following:

```ts
const role = new kplus.Role(this, 'Role');

// configure rules on an imported config map
const cm = kplus.ConfigMap.fromConfigMapName(this, 'ConfigMap', 'config-map-name');
role.allowGet(cm);
```

iliapolo:
Imported resources could not be used in the `role.allowXXX` methods because they didn't implement the `IApiResource` interface. 

They can easily implement this interface because imports are always done using the resource name, which is what this interface needs. 

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/1516

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -16,7 +16,7 @@ export interface ResourceProps {
 /**
  * Represents a resource.
  */
-export interface IResource extends IConstruct {
+export interface IResource extends IConstruct, IApiResource {
   /**
    * The Kubernetes name of this resource.
    */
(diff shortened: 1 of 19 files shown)


src/base.ts
```diff
@@ -27,11 +27,6 @@ export interface IResource extends IConstruct {
    */
   readonly apiVersion: string;
 
-  /**
-   * The group portion of the API version (e.g. "authorization.k8s.io")
-   */
-  readonly apiGroup: string;
-
```
iliapolo, line 34: Removed because its now a duplicate of the same [property](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-25/main/src/api-resource.generated.ts#L10) in `IApiResource`.

vinayak-kukreja (approved):

PR #1576, opened 2023-01-01
feat(container): `optional` in `EnvFrom`

yfried:
Signed-off-by: Yair Fried <5310525+yfried@users.noreply.github.com>


iliapolo:
@yfried Are you looking to pursue this PR? I'm good with it we just need some tests.

PR #1822, opened 2023-03-15
feat(volume): CSI drivers

dancmeyers:
K8s volumes can be created from arbitrary CSI drivers, and can then be mounted on containers in the normal way. This PR adds support for creating CDK8s+ volumes from a CSI driver definition. This includes a generic `attributes` block to match the K8s resource, where driver-specific properties can be set.

The tests in this PR show an example for using the generic 'secrets store' CSI driver, which in turn defines its own CRD for interfacing with a specific provider (e.g. the one to fetch secrets from AWS Secrets Manager or SSM Parameter Store). I would expect this driver/provider to be the most common use-case in an AWS/EKS environment.


src/volume.ts
```diff
@@ -138,6 +139,33 @@ export class Volume extends Construct implements IStorage {
...
+   * The Secret Store CSI driver will need an associated provider to source the
+   * secrets. The following provider allows sourcing from AWS Secrets Manager
+   * and Systems Manager Parameter Store:
+   * https://aws.github.io/secrets-store-csi-driver-provider-aws/
+   *
+   * @param secretProvider The secret provider class to use to populate the volume.
+   * @param options Options
+   */
+  public static fromSecretProvider(
+    scope: Construct, id: string, secretProvider: ss.SecretProviderClass, options: SecretProviderVolumeOptions = { }): Volume {
```
dancmeyers: This will only type match if people use the `SecretProviderClass` from within CDK8s+, which is definitely not ideal. I'm thinking I'll do like the PVC ones and just pass in the name.

src/volume.ts
```diff
@@ -221,6 +221,32 @@ export class Volume extends Construct implements IStorage {
...
+   *
+   * The Secret Store CSI driver will need an associated provider to source the
+   * secrets. The following provider allows sourcing from AWS Secrets Manager
+   * and Systems Manager Parameter Store:
+   * https://aws.github.io/secrets-store-csi-driver-provider-aws/
+   *
+   * @param className The name of the secret provider class to use to populate the volume.
+   * @param options Options
+   */
+  public static fromSecretProviderClass(scope: Construct, id: string, className: string, options: SecretProviderVolumeOptions = { }): Volume {
```
iliapolo: So my concern with this is that we are creating a coupling with specifications we have no control over, that are decoupled from the kubernetes spec itself. 

Imagine that `secrets-store` change either the schema (i.e the `secretProviderClass` property name or structure), or the driver name (i.e `secrets-store.csi.k8s.io`) - what would this function do? it would have to be incompatible with a certain version of the driver. 

I agree that this approach is better that accepting a `SecretProviderClass` object, but the underlying concern still holds true.

This is why integrations with third-party components should be implemented as third-party libraries (ideally published by the vendor itself). The problem of course is that such integrations are currently impossible because the `Volume` class is locked for extensions, but we will address this issue separately. 

In the meantime, I suggest we change this function to support the generic `csi` volume instead. Basically:

```suggestion
  public static fromCsi(scope: Construct, id: string, driver: string, options: CsiVolumeOptions = { }): Volume {
```

Where `CsiVolumeOptions` will contain an `attributes` property to hold the driver specific options.

dancmeyers: Done!

iliapolo (approved):
Sweet, thanks!

dancmeyers:
@mergifyio refresh

PR #1874, opened 2023-03-29
feat(pod): `hostNetwork` property

### LINKED ISSUE #971: `AbstractPodProps` is missing `hostNetwork`

I noticed that when creating a DaemonSet, I'm not able to specify the `hostNetwork: true` part of the pod template spec. There isn't a perfect correspondence between `AbstractPodProps` and `PodSpec`

### Use Case:

I was trying to deploy a DaemonSet which uses the host network.

### Proposed Solution:

Add a `hostNetwork` field to `AbstractPodProps`, corresponding to the one which already exists in `PodSpec`.

Alternatively, I was expecting to be able to specify `DaemonSetSpec` inside of `DaemonSetProps`, but for some reason the spec is only generated by an internal function and can't be provided to the constructor. I guess you have a historical reason for this, but it seems undesirable to me.

---

sumupitchayan:
This PR fixes the above issue by adding the `hostNetwork` property to `AbstractPodProps`.

Changes were tested by adding 2 new test cases for the `hostNetwork` property and updating the existing test snapshots to include this new property.

Resolves [#971](https://github.com/cdk8s-team/cdk8s/issues/971)


iliapolo (approved):

aarashy:
Cool, thank you!

PR #1893, opened 2023-04-14
feat: Exposing readinessGates to high level objects on AbstractPod

RustyChain:
feat: Exposing readinessGates to high-level objects (Workload and Deployments classes) on AbstractPod.

As per [AWS Ingress Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v1.1/guide/ingress/pod-conditions/) documentation says, readinessGates are crucial for zero downtime deployments.

### DIFF
--- a/src/pod.ts
+++ b/src/pod.ts
@@ -26,6 +26,7 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
   private readonly _initContainers: container.Container[] = [];
   private readonly _hostAliases: HostAlias[] = [];
   private readonly _volumes: Map<string, volume.Volume> = new Map();
+  private readonly _readinessGates: k8s.PodReadinessGate[] = [];
 
   public abstract readonly podMetadata: ApiObjectMetadataDefinition;
 
@@ -56,6 +57,10 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
       props.hostAliases.forEach(c => this.addHostAlias(c));
     }
 
+    if (props.readinessGates) {
+      props.readinessGates.forEach(c => this.addReadinessGate(c));
+    }
+
   }
 
   public get containers(): container.Container[] {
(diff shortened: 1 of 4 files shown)


RustyChain:
If the pr will get merged, then I'll raise another pr for v1.25/v1.26 branches to add the changes for the actual versions

src/pod.ts
```diff
@@ -423,6 +442,12 @@ export interface AbstractPodProps extends base.ResourceProps {
    * @default false
    */
   readonly isolate?: boolean;
+  /**
+   * Creates readinessGate for the ingress management
+   *
+   * @default k8s.PodReadinessGate[]
+   */
+  readonly readinessGates?: k8s.PodReadinessGate[];
```
iliapolo, line 450: We don't expose `k8s.*` interfaces or resource from the L2 layer, as this creates a coupling between the L1 and L2 API's, which is undesired. 

Instead. create an interface in this file called `PodRedinessGate` which includes whatever properties you want to expose. We prefer some duplication over creating the aforementioned coupling. 

Also, looking at the available properties on this struct, I see:

```ts
  /**
   * ConditionType refers to a condition in the pod's condition list with matching type.
   *
   * @schema io.k8s.api.core.v1.PodReadinessGate#conditionType
   */
  readonly conditionType: string;
```

What are the values that `conditionType` can accept? Are these really arbitrary strings or can we wrap it with a type safe API? This kind of stuff is exactly where cdk8s-plus shines and can provide a much simpler experience over just using the L1 layer (or YAML for that matter).

iliapolo (commented):
Hi @RustyChain.

Thanks for your work on this. See my comments. Also, please rebase this PR to target the `k8s-26/main` branch, `k8s-22/main` is not supported anymore.

Also, no need to create additional PR's. If this gets merged into `k8s-26/main`, it will be automatically backported to `k8s-25/main` and `k8s-24/main`, which are the three versions currently supported.

PR #1894, opened 2023-04-16
chore: decouple optional and required properties for a container

msessa:
This simple PR decouples the optional properties of a container from the required ones.

Rationale:

We use this library in combination with CDK's capabilities to handle compiling, bundling and uploading images to ECR. In which case we expect our users to provide all other properties of a container but the `image` property which is abstracted away.

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -494,13 +494,18 @@ export interface ContainerLifecycle {
 /**
  * Properties for creating a container.
  */
-export interface ContainerProps {
+export interface ContainerProps extends ContainerOpts {
 
   /**
    * Docker image name.
    */
   readonly image: string;
+}
 
+/**
+ * Optional properties of a container.
+ */
+export interface ContainerOpts {
   /**
    * Name of the container specified as a DNS_LABEL. Each container in a pod must have a unique name (DNS_LABEL). Cannot be updated.
    *


msessa:
@iliapolo any eta for a review?

iliapolo (commented):
@msessa Not sure I understand what this PR accomplishes. 

Technically, this PR doesn't change any API because I don't see `ContainerOpts` being used anywhere. Can you please elaborate on how you see this being used? Preferable in the formt of a "futuristic" code snippet.

msessa:
Hi @iliapolo,

`ContainerOpts` can be used by a downstream library to provide high-level constructs that handle the lifecycle of a docker image, therefore the end-user will not need to pass `image` as a property but might still be configuring other aspects of the container (eg, resource allocations, env vars, etc)

Take this example taken from one of our internal libraries.

The construct handles all the complexity of producing an asset directory with a Dockerfile, and then uses CDK's capabilities of uploading the asset to an ECR registry.
A user of our library still might want to configure other properties of the container.

As you can see it'd be confusing to require a `ContainerProps` since the user will be forced to pass something for `image`

```typescript
/**
 * Properties for DotnetDeploymentApplication
 */
export interface DotnetDeploymentApplicationProps {
  readonly containerOpts?: ContainerOpts;
}

/**
 * Creates a single-deployment dotnet application
 */
export class DotnetDeploymentApplication extends Construct {
  /**
   * The image asset location used by the deployment's main container
   */
  public readonly imageAssetLocation: DockerImageAssetLocation;

  constructor(scope: Construct, id: string, props: DotnetDeploymentApplicationProps) {
    super(scope, id);
    const chart = new Chart(new App(), 'App Chart');

    // This pseudo-function compile and bundles some code, and creates a CDK image asset
    // returning a `DockerImageAssetLocation` class
    // CDK will then handle the creation of the docker image and uploading to ECR.
    // The image location will then be known at deploy-time
    const image = DotnetBundling.getImageLocation(this);

    // Provision a deployment
    const deployment = new kplus.Deployment(chart, 'Deployment');

    // End-user provides all properties of the main container except for the image property
    deployment.addContainer({
      image: image.imageUri,
      ...props.containerOpts,
    })

    // ....some code to deploy the chart (eg `cdk.aws_eks.Cluster.addCdk8sChart()`)
  }
}
```

iliapolo (approved):
@msessa Thanks for details 👍

PR #2026, opened 2023-05-22
feat(pod): termination grace period

### LINKED ISSUE #1413: Support for terminationGracePeriodSeconds
Is it on the roadmap to add support for this field on Deployments?

galsasi1989:
Add option to configure grace period in seconds until the pod is terminated 

Resolves https://github.com/cdk8s-team/cdk8s-plus/issues/1413


src/pod.ts
```diff
@@ -437,6 +440,13 @@ export interface AbstractPodProps extends base.ResourceProps {
    * @default false
    */
   readonly hostNetwork?: boolean;
+
+  /**
+   * Grace period until the pod is terminated
+   *
+   * @default Duration.seconds(30)
+   */
+  readonly terminationGracePeriodSeconds?: Duration;
```
iliapolo: It doesn't have to be seconds actually, `Duration` has a other APIs :)

```suggestion
  readonly terminationGracePeriod?: Duration;
```

iliapolo (changes requested):
Sweet, thanks!

Notice the DCO check is failing, you can fix it by following the steps [here](https://github.com/cdk8s-team/cdk8s-plus/pull/2026/checks?check_run_id=13684600443).
And to prevent it from happening in the future, see [DCO](https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-26/main/CONTRIBUTING.md#developer-certificate-of-origin-dco).

src/pod.ts
galsasi1989: According to k8s api, it's in seconds: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination:~:text=terminationGracePeriodSeconds
Unless I misunderstood :)

iliapolo: Yes, eventually you'll have to convert it to seconds (like you already are). But a user of cdk8s-plus can pass any `Duration` object. For example:

```ts
terminationGracePeriodSeconds: Duration.hours(1)`
```

This is great, cdk8s-plus offers a better API then k8s because it does the conversion for you.
So a better name to the property would be just `terminationGracePeriod`.

galsasi1989: @iliapolo - I changed it to terminationGracePeriod and updated the tests accordingly

iliapolo (approved):
🚢 🚢

PR #2302, opened 2023-07-14
feat(volume): nfs volume support (#2300)

### LINKED ISSUE #2300: Add ability to create NFS Volumes
NFS Volumes

### Use Case:
I need to mount NFS volumes to my pods. Unfortunately, due to how the `Volume` class is set up, I cannot easily create an NFS volume to pass into the `Deployment` construct.

My current workaround for this is the following:
```
const volume = Volume.fromEmptyDir(this, 'data-volume', 'data-volume', { });
const obj: any = volume as any;
obj.config = {
    nfs: {
        server: '169.254.0.1'
    }
};
```

### Proposed Solution:
 - Add a new static constructor on the `Volume` class to support creating NFS volumes. The underlying `k8s.Volume` already has `nfs` property defined on it.
 - To support other volume types that do not have static constructors defined, I suggest making the `Volume` class constructor public or protected, so that users are left with flexibility to pass in their own configuration objects if they have to.

### Other:
I can help put together a PR if needed

* [X] :wave: I may be able to implement this feature request

---

wyvernzora:
Fixes #2300


iliapolo (approved):
Sweet, thanks!

PR #2964, opened 2023-09-25
fix(ingress): add `ingressClassName` to `IngressProps`

### LINKED ISSUE #2856: Add IngressClassName to Ingress
We would like to be able to configure the `IngressClassName` on `Ingress` objects. 

### Use Case:
The IngressClassName is part of the officiall documentation and was added within Kubernetes version 1.18.

"used to reference the name of the Ingress controller that should implement the Ingress, the field is a reference to an IngressClass resource that contains additional Ingress configuration, including the name of the Ingress controller" (https://kubernetes.io/docs/concepts/services-networking/ingress)

### Proposed Solution:
Currently we have implemented it as an extension like follows:
``` TypeScript
import { JsonPatch } from "cdk8s";
import { Ingress, IngressProps } from "cdk8s-plus-25";
import { Construct } from "constructs";

export class IngressWithClassName extends Ingress {
    constructor(scope: Construct, id: string, props: IngressProps & { ingressClassName: string }) {
        super(scope, id, props);

        this.apiObject.addJsonPatch(
            JsonPatch.add(
                '/spec/ingressClassName',
                props.ingressClassName
            )
        );
    }
}
```

---

sumupitchayan:
Fixes #2856 

Currently, `IngressProps` does not have a property allowing users to pass in an `ingressClassName`. This PR fixes this and adds a unit test to verify the change.


src/ingress.ts
```diff
@@ -40,6 +40,11 @@ export interface IngressProps extends base.ResourceProps {
    * extension, if the ingress controller fulfilling the ingress supports SNI.
    */
   readonly tls?: IngressTls[];
+
+  /**
+   * Ingress Class Name is an optional String property.
+   */
+  readonly ingressClassName?: string;
```
iliapolo: "ingress" is already implied 

```suggestion
  readonly className?: string;
```

src/ingress.ts
```diff
@@ -40,6 +40,11 @@ export interface IngressProps extends base.ResourceProps {
    * extension, if the ingress controller fulfilling the ingress supports SNI.
    */
   readonly tls?: IngressTls[];
+
+  /**
+   * Ingress Class Name is an optional String property.
```
iliapolo: Describe the meaning, not the technicality. 

That is, what does `ingressClassName` mean? We usually lean on the k8s docs for that.

iliapolo (changes requested):

pivstone:
Hi there, looks like this PR is stale for quite some time.  
I'm wondering what's the next step for this PR? 

It will be nice we could implement it. We are moving our helm to use CDK8S, and we found CDK8S don't support `ingressClassName`. We also saw the work around. But it will be nice if we can support it directly,

iliapolo (approved):

PR #3508, opened 2024-01-01
fix(container): allow the same port number with different protocols

### LINKED ISSUE #3507: Support multi-protocol ports on containers
### Description of the bug:
There are some situations where the same port on a container can serve multiple protocols. One example being DNS servers typically serving both TCP and UDP traffic from port 53. The `Container` construct, however, fails when a port number repeats, even if the specified protocol is different.

There are examples of specifying the same port multiple times with different protocols. One below is an example of a `docker-compose.yml` file where this happens:
```
version: "2.1"
services:
  blocky:
    image: spx01/blocky
    container_name: blocky
    restart: unless-stopped
    # Optional the instance hostname for logging purpose
    hostname: blocky-hostname
    ports:
      - "53:53/tcp"
      - "53:53/udp"
      - "4000:4000/tcp"
    environment:
      - TZ=Europe/Berlin # Optional to synchronize the log timestamp with host
    volumes:
      # Optional to synchronize the log timestamp with host
      - /etc/localtime:/etc/localtime:ro
      # config file
      - ./config.yml:/app/config.yml
```

### Reproduction Steps:
Synth the following `Deployment` in any context
```
new Deployment(app, 'test-depl', {
    containers: [{
        name: 'blocky',
        ports: [{
                name: 'dns-tcp',
                number: 53,
                protocol: Protocol.TCP,
            }, {
                name: 'dns-udp',
                number: 53,
                protocol: Protocol.UDP,
            }, {
                name: "http",
                number: 4000,
                protocol: Protocol.TCP,
            }],
    }],
});
```

### Error Log:
As of `cdk8s-plus-27@2.7.70` trying to specify the same port number twice results in the following error:
```
Error: Port with number 53 already exists
    at Container.addPort (/Users/wyvernzora/Code/k2/gitops/infra/dns/node_modules/cdk8s-plus-27/src/container.ts:830:13)
    at new Container (/Users/wyvernzora/Code/k2/gitops/infra/dns/node_modules/cdk8s-plus-27/src/container.ts:770:12)
   ...
```

### Environment:

  - **Framework Version**: 
      - `cdk8s-cli@2.198.21`
      - `cdk8s@2.68.22`
      - `cdk8s-plus-27@2.7.70`
  - **OS**: MacOS Sonoma 14.2

### Other:
The culprit looks like [this comparison here](https://github.com/cdk8s-team/cdk8s-plus/blob/a1b57f55959771e9490d893545771dde9339df88/src/container.ts#L829). I think the protocol should also be factored in here. Or, alternatively, allow specifying multiple protocols per ContainerPort.

---

This is :bug: Bug Report

wyvernzora:
Fixes #3507


iliapolo (approved):
👍

PR #3672, opened 2024-02-08
feat(pod-spec): add shareProcessNamespace to k8s-26

chensjlv:
add shareProcessNamespace to pod spec. Cannot find it anywhere.

https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/


iliapolo:
Not needed. Backports happen automatically.

PR #3812, opened 2024-03-01
feat(container): support `restartPolicy` field

serhatcetinkaya:
[KEP-753](https://github.com/kubernetes/enhancements/tree/e1a916745378660bafb96cb3c9f754fa793342f9/keps/sig-node/753-sidecar-containers) introduced native sidecar container support starting from kubernetes v1.28

This PR adds the field to Container object. Existing tests are working, I added one more test to make sure field is there if the property is specified


src/container.ts
```diff
@@ -658,6 +679,15 @@ export interface ContainerOpts {
    *   group: 26000
    */
   readonly securityContext?: ContainerSecurityContextProps;
+
+  /**
+   * Kubelet will start init containers with restartPolicy=Always in the order with other init containers,
+   * but instead of waiting for its completion, it will wait for the container startup completion
+   * Currently, only accepted value is Always
+   * @see https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/
+   * @default - no restart policy is defined
```
iliapolo:
```suggestion
   * @default - no restart policy is defined and the pod restart policy is applied
```

src/container.ts
```diff
@@ -491,6 +491,27 @@ export interface ContainerLifecycle {
 
 }
 
+/**
+ * RestartPolicy defines the restart behavior of individual containers in a pod.
+ * This field may only be set for init containers, and the only allowed value is "Always".
```
iliapolo, line 496: How about adding a validation that normal containers don't specify this? We can do it here: 

https://github.com/cdk8s-team/cdk8s-plus/blob/0cd21c04e533e873c1698d173d395e4338e2a8ec/src/pod.ts#L192-L199

test/container.test.ts
```diff
@@ -473,6 +473,26 @@ describe('Container', () => {
...
+      image: 'bar',
+      restartPolicy: ContainerRestartPolicy.ALWAYS,
+    });
+
+    // THEN
+    const manifest = Testing.synth(chart);
+    expect(manifest).toMatchSnapshot();
+    const container = manifest[0].spec.initContainers[0];
+
+    expect(container).toHaveProperty('restartPolicy');
```
iliapolo:
```suggestion
    expect(container.restartPolicy).toEqual('Always');
```

iliapolo (changes requested):
Thanks for this! 

Can you also make sure the PR [allows changes from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)? I can't seem to update your branch from origin. Otherwise you'll have to manually update (which is also ok if you want)

serhatcetinkaya:
@iliapolo thanks for the detailed check, I added all your suggestions. It should be ok now, can you give it another look please?

iliapolo (approved):

PR #3813, opened 2024-03-01
fix(role): use rules coming from the roleProps

serhatcetinkaya:
This fixes a bug where any rule given from roleProps is not used. the for loop exists for clusterRole but was missing on Role. a test is included to prevent breaking in future


iliapolo (approved):
Nice!

PR #3818, opened 2024-03-01
feat(probe): add `host` property to HttpGet probes

serhatcetinkaya:
HttpGet probes has `host` field. Added support for it in cdk8s-plus:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#httpgetaction-v1-core


iliapolo (approved):
👍

PR #4174, opened 2024-05-17
feat: add ability to add ConfigMap checksum as resource annotations

wyvernzora:
Note: draft change do not merge

One common practice that I have seen is adding the checksum of config map contents as annotations to workloads, so that when a config map is updated, the associated deployments can also get updated and bounced. This PR is an attempt to implement such functionality, though I am not sure about some aspects of this change and would appreciate some feedback.

Specific things that I'd like some feedback on:
 - Should we expose the checksum as a property on ConfigMap? For example `cm.checksum`. While the type of that property would need to be `string`, but in fact it needs to be `Lazy` that produces a string. I am not sure what the practice is for exposing a `Lazy` as a property on a construct.
 - Choosing to not expose the checksum outside the `ConfigMap` results in this awkward `addChecksumTo()` API. I'd appreciate any input on what would be a good API pattern for adding these checksum annotations.
 - I had to add a dependency on `json-stable-stringify` to make sure that hashes come out consistent. What are the opinions of this project on adding dependencies?


src/config-map.ts
```diff
@@ -227,9 +225,25 @@ export class ConfigMap extends base.Resource implements IConfigMap {
...
   }
 
+  /**
+   * Adds an annotation to the supplied object that contains the checksum of the
+   * config map contents.
+   * @param obj Resource object to add the annotation to
+   */
+  public addChecksumTo(obj: base.Resource) {
+    obj.metadata.addAnnotation(
+      `checksum/${this.node.id}`,
```
iliapolo, line 235: `node.id` is unique only in a given scope, `node.addr` will give an app unique id.
 
```suggestion
      `cdk8s.io/config-map.${this.node.addr}.checksum`,
```

test/config-map.test.ts
```diff
@@ -317,5 +313,142 @@ test('can configure an immutable config map', () => {
...
+
+  // WHEN
+  const cm = new ConfigMap(chart, 'my-config-map', {
+    data: {
+      key1: 'foo',
+      key2: 'bar',
+    },
+  });
+  const ns = new Namespace(chart, 'my-namespace', {});
+  cm.addChecksumTo(ns);
```
iliapolo, line 330: Please add documentation for this feature to https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-30/main/docs/plus/config-map.md

iliapolo:
@wyvernzora 

> Should we expose the checksum as a property on ConfigMap? 

No need. At least not right now.

> Choosing to not expose the checksum outside the ConfigMap results in this awkward addChecksumTo()

Agree. But its ok. Exposing lazy strings is also quite awkward. 

> I had to add a dependency on json-stable-stringify to make sure that hashes come out consistent. What are the opinions of this project on adding dependencies?

We try to avoid it, but also recognize sometimes it is needed. This seems like one of those times 👍

burner1024:
Please reconsider.

PR #3817, opened 2024-03-01
feat(container): add capabilities to security context

serhatcetinkaya:
adding support for capabilities in container security context. see:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#capabilities-v1-core

PR #4267, opened 2024-06-05
fix(container): remove CAP_ prefix from capabilities

serhatcetinkaya:
looks like I made a mistake in #3817 

in POSIX capabilities list we have `CAP_` prefix, but in kubernetes container spec we omit it. 

from official [docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/):

> Linux capability constants have the form CAP_XXX. But when you list capabilities in your container manifest, you must omit the CAP_ portion of the constant. For example, to add CAP_SYS_TIME, include SYS_TIME in your list of capabilities.

### DIFF
--- a/src/container.ts
+++ b/src/container.ts
@@ -18,167 +18,167 @@ export enum Capability {
   /**
    * CAP_AUDIT_CONTROL
    */
-  AUDIT_CONTROL = 'CAP_AUDIT_CONTROL',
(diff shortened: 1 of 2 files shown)


iliapolo (approved):

iliapolo:
@Mergifyio refresh

PR #4483, opened 2024-08-05
fix(pod): allow sidecar containers to have probes

### LINKED ISSUE #4338: Sidecar containers cannot specify probes
Sidecar containers, unlike init containers, [support probes](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#differences-from-init-containers). `addInitContainer` doesn't support this and will throw an error if `liveness`, `readiness`, or `startup` are specified:
https://github.com/cdk8s-team/cdk8s-plus/blob/f7c11148f65084b35fb5d3273a31af387015bc8f/src/pod.ts#L124-L135

I'd like to be able to specify these probes for sidecar containers. 

### Use Case:
Since a sidecar container does not exit before the main container starts, probes are the only means available for delaying the startup of the main container beyond the simplest possible cases. Sidecar containers may require some time to be ready after starting, and in my case it's important that the sidecar container be _ready_, not just _started_, by the time the main container starts.

### Proposed Solution:

It seems relatively straightforward to build on #3812 and disable the exceptions if `restartPolicy` is specified.

### Other:

---

serhatcetinkaya:
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#differences-from-regular-containers

if we check docs, sidecar containers support readiness, liveness and startup probes. I added an extra check to make sure the init container is not sidecar for probe checks.

this way it will be possible to add probes for initcontainers when they are sidecar.

edit: this should be backported to k8s-29/main as well

Fixes #4338


serhatcetinkaya:
I noticed this fixes #4338

iliapolo:
> this should be backported to k8s-29/main as well

@serhatcetinkaya what about k8s-28? no need?

serhatcetinkaya:
my mistake, I thought #3812 was not backported to k8s-28. we should backport to that branch as well. thanks for the notice

src/pod.ts
```diff
@@ -122,15 +122,15 @@ export abstract class AbstractPod extends base.Resource implements IPodSelector,
   public addInitContainer(cont: container.ContainerProps): container.Container {
 
     // https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#differences-from-regular-containers
-    if (cont.readiness) {
+    if (cont.restartPolicy != container.ContainerRestartPolicy.ALWAYS && cont.readiness) {
```
iliapolo: So a container that has `ContainerRestartPolicy.ALWAYS` is considered a sidecar? Can you elaborate and refer to some docs explaining this? Also, lets put the new condition in a function:

```suggestion
    if (!this.isSidecar(cont) && cont.readiness) {
```

iliapolo: We also need some tests for these new conditions

iliapolo (changes requested):

src/pod.ts
iliapolo: Ah ok I see its explained in the issue: https://github.com/cdk8s-team/cdk8s-plus/issues/4338#issuecomment-2191018002.

I would still make a note on this in the code for future readers, and put it in a separate function so its more readable.

serhatcetinkaya: > So a container that has ContainerRestartPolicy.ALWAYS is considered a sidecar? Can you elaborate and refer to some docs explaining this?

basically, yes. any init container having `ContainerRestartPolicy.ALWAYS` is a sidecar container. you can see the differences here: https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/#differences-from-init-containers

I see your concerns. I will add a new function to check if container is sidecar. Also I will add a comment with a link to documentation for future collaborators and add more tests and try to cover everything

serhatcetinkaya:
@iliapolo hello again, all the suggestions are implemented. Can you take another look when you have time?

iliapolo (approved):
Perfect. Thank you!

PR #4902, opened 2024-11-13
feat(pod): shareProcessNamespace

chensjlv:
add shareProcessNamespace to pod spec. Cannot find it anywhere.

https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/


iliapolo (approved):
Thanks!

PR #5469, opened 2025-03-18
feat: Implement `Service.fromServiceName()`

### LINKED ISSUE #2180: Implement `Service.fromServiceName`
More often than not, there is a need to re-use an existing `Service` not created or managed by `cdk8s`. Elsewhere, you can re-use an existing `ServiceAccount` by calling `ServiceAccount.fromServiceAccountName`, but that functionality is not available for `Service`s. 

---

rassie:
This is my first stab on extending `cdk8s`, so please review thoroughly.  

Fixes: cdk8s-team/cdk8s#2180

### DIFF
--- a/src/ingress.ts
+++ b/src/ingress.ts
@@ -270,9 +270,20 @@ export class IngressBackend {
    * A Kubernetes `Service` to use as the backend for this path.
    * @param serv The service object.
    */
-  public static fromService(serv: service.Service, options: ServiceIngressBackendOptions = {}) {
-    if (serv.ports.length === 0) {
-      throw new Error('service does not expose any ports');
+  public static fromService(serv: service.IService, options: ServiceIngressBackendOptions = {}) {
+    // For imported services, ports might be undefined or empty
+    // In that case, we must rely on the port provided in options
+    if (!serv.ports || serv.ports.length === 0) {
+      if (options.port !== undefined) {
+        return new IngressBackend({
+          service: {
+            name: serv.name,
+            port: { number: options.port },
+          },
+        });
+      } else {
+        throw new Error('service does not expose any ports');
+      }
     }
 
     let servicePort;
--- a/src/service.ts
+++ b/src/service.ts
@@ -7,6 +7,42 @@ import * as ingress from './ingress';
 import * as pod from './pod';
 
 /**
+ * Interface for Service constructs.
+ * This interface allows users to interact with a Service without having to create one directly.
+ */
+export interface IService extends base.IResource {
+  /**
+   * The ports this service binds to.
+   */
+  readonly ports: ServicePort[];
+
+  /**
+   * The first port of the service, if available.
+   */
+  readonly port?: number;
+
+  /**
+   * Expose a service via an ingress using the specified path.
+   *
+   * @param path The path to expose the service under.
+   * @param options Additional options.
+   *
+   * @returns The `Ingress` resource that was used.
+   */
+  exposeViaIngress(path: string, options?: ExposeServiceViaIngressOptions): ingress.Ingress;
+
+  /**
+   * Require this service to select pods matching the selector.
+   *
+   * Note that invoking this method multiple times acts as an AND operator
+   * on the resulting labels.
+   *
+   * @param selector The pod selector
+   */
+  select(selector: pod.IPodSelector): void;
+}
+
+/**
  * Properties for `Service`.
  */
 export interface ServiceProps extends base.ResourceProps {
@@ -124,6 +160,14 @@ export interface ExposeServiceViaIngressOptions {
    * @default - An ingress will be automatically created.
    */
   readonly ingress?: ingress.Ingress;
+
+  /**
+   * The port to use for the service backend.
+   *
+   * @default - For regular services, the first port or port 80 will be used.
+   *            For imported services, this is required.
+   */
+  readonly port?: number;
 }
 
 /**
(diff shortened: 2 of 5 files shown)


src/ingress.ts
```diff
@@ -270,9 +270,20 @@ export class IngressBackend {
...
-      throw new Error('service does not expose any ports');
+  public static fromService(serv: service.IService, options: ServiceIngressBackendOptions = {}) {
+    // For imported services, ports might be undefined or empty
+    // In that case, we must rely on the port provided in options
+    if (!serv.ports || serv.ports.length === 0) {
+      if (options.port !== undefined) {
+        return new IngressBackend({
+          service: {
+            name: serv.name,
+            port: { number: options.port },
```
iliapolo, line 281: I understand where you're going with this, but it causes `options.port` to have multiple designations in the function. Note that `options.port` is currently only used as _disambiguator_ in case the service itself exposes more than one port, as evident by:

https://github.com/cdk8s-team/cdk8s-plus/blob/5963ee2714d1f45b9cb7fa738f93afcbb6207b8d/src/ingress.ts#L294-L296

With this change though, it suddenly becomes an _override_, just for this specific code path. 

Instead, I prefer we allow the user to specify a port during service import. See below.

src/service.ts
```diff
@@ -177,6 +221,83 @@ export interface AddDeploymentOptions extends ServiceBindOptions {
...
+
+  /**
+   * For imported services, we don't support selecting pods.
+   * This is a no-op for compatibility with the IService interface.
+   *
+   * @param _selector The pod selector (unused)
+   */
+  public select(_selector: pod.IPodSelector): void {
+    // No-op for imported services
+  }
```
iliapolo, line 278: Im not too groovy on all these no-ops/empty values. Users calling these will either not know how to handle it, or mistakenly think the operation did what they wanted. Interface functionality that cannot be implemented on an imported resource, should throw an error:

```ts
public select(_selector: pod.IPodSelector): void {
    throw new Error('Pod selection is not possible with imported services');
  }
```

Which is why we normally create interfaces with a very slim surface area. In this case, I believe the `IService` interface should be reduced to only include functionality that **can** be implemented.

src/service.ts
```diff
@@ -177,6 +221,83 @@ export interface AddDeploymentOptions extends ServiceBindOptions {
...
+   *
+   * @returns The `Ingress` resource that was used.
+   */
+  public exposeViaIngress(path: string, options: ExposeServiceViaIngressOptions = {}): ingress.Ingress {
+    const ingr = options.ingress ?? new ingress.Ingress(this, 'Ingress');
+
+    // For imported services, we require a port to be specified in the options
+    if (!options.port) {
+      throw new Error('A port must be specified when exposing an imported service via ingress');
+    }
```
iliapolo, line 294: If the imported service would allow specifying ports, this would not be needed, and the implementation could be shared between `Service` and `ImportedService`

test/service.test.ts
```diff
@@ -196,3 +196,14 @@ test('can set publishNotReadyAddresses', () => {
...
+test('can be imported and used by an ingress', () => {
+  const chart = Testing.chart();
+
+  const service = kplus.Service.fromServiceName(chart, 'Service', 'service-name');
+
+  // Now we must specify a port for imported services
+  service.exposeViaIngress('/hello', { port: 8080 });
+  const ingress = Testing.synth(chart)[0];
+  expect(ingress).toMatchSnapshot();
+});
```
iliapolo, line 209: We should change the implementation to allow for the following API:

```suggestion
test('can be imported and used by an ingress', () => {
  const chart = Testing.chart();

  // import an existing service that exposes port 8000 and is called 'service-name'. 
  const service = kplus.Service.fromServiceName(chart, 'Service', 'service-name', { port: 8000 });

  // same api as with a managed service
  service.exposeViaIngress('/hello');
  const ingress = Testing.synth(chart)[0];
  expect(ingress).toMatchSnapshot();
});
```

src/service.ts
```diff
@@ -6,6 +6,42 @@ import * as k8s from './imports/k8s';
...
+
+  /**
+   * Require this service to select pods matching the selector.
+   *
+   * Note that invoking this method multiple times acts as an AND operator
+   * on the resulting labels.
+   *
+   * @param selector The pod selector
+   */
+  select(selector: pod.IPodSelector): void;
```
iliapolo, line 42: Why did you add this to the interface if it cannot be implemented by imported services?

src/service.ts
```diff
@@ -6,6 +6,42 @@ import * as k8s from './imports/k8s';
...
+export interface IService extends base.IResource {
+  /**
+   * The ports this service binds to.
+   */
+  readonly ports: ServicePort[];
+
+  /**
+   * The first port of the service, if available.
+   */
+  readonly port?: number;
```
iliapolo, line 22: Do we really need this on the interface? From what I could tell, just `ports` is needed. Actually the fact both `port` and `ports` even exist is for backward compat; but the new interface doesn't have to suffer from this.

src/service.ts
```diff
@@ -177,6 +221,83 @@ export interface AddDeploymentOptions extends ServiceBindOptions {
...
+  public get resourceName(): string {
+    return this.name;
+  }
+
+  /**
+   * For imported services, we don't have information about the ports.
+   * If specific port information is needed, it should be passed in configuration.
+   */
+  public get ports(): ServicePort[] {
+    return [];
```
iliapolo, line 259: With the new proposed API, this will return the user provided ports (if any):

```suggestion
    return [...this._ports];
```

Exactly like the managed service does.

iliapolo (changes requested):

rassie:
@iliapolo I've been a bit swamped, will take care of the comments some time next week.

PR #5470, opened 2025-03-18
feat(stateful-set): support `PersistentVolumeClaimTemplates`

### LINKED ISSUE #2183: Support PersistentVolumeClaimTemplate in STS

STS supports PVC templates and cdk8s should do

### Use Case:

To deploy multiple pods with storage

### Proposed Solution:

Update StatefulSetProps to allow PVC template

### Other:

---

rassie:
I've tried adding support for template claims. Please review gently but thoroughly.

Fixes: cdk8s-team/cdk8s#2183

### DIFF
--- a/docs/plus/volume.md
+++ b/docs/plus/volume.md
@@ -36,3 +36,39 @@ const redis = pod.addContainer({
 // mount to the redis container.
 redis.mount('/var/lib/redis', data);
 ```
+
+## Using PersistentVolumeClaim Templates with StatefulSets
+
+When working with StatefulSets, you can use PersistentVolumeClaim templates to create stable storage for each pod in the StatefulSet. This allows each pod to have its own storage that persists even if the pod is rescheduled to a different node.
+
+```typescript
+import * as kplus from 'cdk8s-plus-32';
+import { Size } from 'cdk8s';
+
+// Create a StatefulSet with a PVC template
+const statefulSet = new kplus.StatefulSet(this, 'StatefulSet', {
+  containers: [{
+    image: 'nginx',
+    volumeMounts: [{
+      path: '/data',
+      name: Volume.fromName(chart, "pvc-template", "data-volume"),
+    }],
+  }],
+  // Define PVC templates during initialization
+  volumeClaimTemplates: [{
+    name: 'data-volume',  // Must match a volume mount name in a container
+    storage: Size.gibibytes(10),
+    accessModes: [kplus.PersistentVolumeAccessMode.READ_WRITE_ONCE],
+    storageClassName: 'standard',  // Optional: Specify the storage class
+  }],
+});
+
+// Or add PVC templates after creation
+statefulSet.addVolumeClaimTemplate({
+  name: 'logs-volume',
+  storage: Size.gibibytes(5),
+  accessModes: [kplus.PersistentVolumeAccessMode.READ_WRITE_ONCE],
+});
+```
+
+Each pod in the StatefulSet will get its own PVC instance based on these templates, with names like `data-volume-my-statefulset-0`, `data-volume-my-statefulset-1`, etc.
--- a/src/stateful-set.ts
+++ b/src/stateful-set.ts
@@ -3,6 +3,7 @@ import { Construct } from 'constructs';
 import * as container from './container';
 import { IScalable, ScalingTarget } from './horizontal-pod-autoscaler';
 import * as k8s from './imports/k8s';
+import * as pvc from './pvc';
 import * as service from './service';
 import * as workload from './workload';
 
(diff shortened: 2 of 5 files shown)


src/stateful-set.ts
```diff
@@ -93,6 +104,19 @@ export interface StatefulSetProps extends workload.WorkloadProps {
  * - Ordered, graceful deployment and scaling.
  * - Ordered, automated rolling updates.
  */
+/**
+ * A PersistentVolumeClaim template for StatefulSets
+ */
+export interface PersistentVolumeClaimTemplate extends pvc.PersistentVolumeClaimProps {
```
iliapolo, line 110:
```suggestion
export interface PersistentVolumeClaimTemplateProps extends pvc.PersistentVolumeClaimProps {
```

src/stateful-set.ts
```diff
@@ -93,6 +104,19 @@ export interface StatefulSetProps extends workload.WorkloadProps {
  * - Ordered, graceful deployment and scaling.
  * - Ordered, automated rolling updates.
  */
+/**
+ * A PersistentVolumeClaim template for StatefulSets
+ */
```
iliapolo, line 109: This cuts off the doc string for `StatefulSet`, needs to moved up along with `PersistentVolumeClaimTemplate`

docs/plus/volume.md
```diff
@@ -36,3 +36,39 @@ const redis = pod.addContainer({
...
+  containers: [{
+    image: 'nginx',
+    volumeMounts: [{
+      path: '/data',
+      name: Volume.fromName(chart, "pvc-template", "data-volume"),
+    }],
+  }],
+  // Define PVC templates during initialization
+  volumeClaimTemplates: [{
+    name: 'data-volume',  // Must match a volume mount name in a container
```
iliapolo, line 59: The volume mount name is actually always the name of the volume:

https://github.com/cdk8s-team/cdk8s-plus/blob/5963ee2714d1f45b9cb7fa738f93afcbb6207b8d/src/container.ts#L1120-L1122

So, can we not accept a `Volume` here instead of a `string`? Then extract the name automatically? 
The thing that confuses me though is that `volume` is an optional property in `PersistentVolumeClaimProps`:

https://github.com/cdk8s-team/cdk8s-plus/blob/5963ee2714d1f45b9cb7fa738f93afcbb6207b8d/src/pvc.ts#L54-L70

So what will happen if a volume is not provided? 😕

src/stateful-set.ts
```diff
@@ -93,6 +104,19 @@ export interface StatefulSetProps extends workload.WorkloadProps {
...
+export interface PersistentVolumeClaimTemplate extends pvc.PersistentVolumeClaimProps {
+  /**
+   * The name of the claim that the StatefulSet controller will create for each pod.
+   * This will be used to name the created PVC in the format <claim-name>-<pod-name>
+   *
+   * This name should match the name of a volume mount in one of the containers.
```
iliapolo, line 115: What happens if no matching volume mount is found? should we throw an error?

src/stateful-set.ts
```diff
@@ -178,21 +208,76 @@ export class StatefulSet extends workload.Workload implements IScalable {
...
+    // Get base pod spec
+    const basePodSpec = this._toPodSpec();
+
+    // Create a new pod spec for the StatefulSet with filtered volumes
+    const podSpec = { ...basePodSpec };
+
+    // When using volumeClaimTemplates, the volumes with matching names should not be included in the pod spec
+    if (this._volumeClaimTemplates.length > 0 && podSpec.volumes) {
+      const volumeClaimNames = this._volumeClaimTemplates.map(vct => vct.name);
+      const filteredVolumes = podSpec.volumes.filter(vol => !volumeClaimNames.includes(vol.name));
```
iliapolo, line 243: I think is missing a test case, where we define one mount that has a corresponding template, and one that doesn't.

iliapolo, line 243: Give more prescriptive names:

```suggestion
      const volumesWithNoTemplates = podSpec.volumes.filter(vol => !volumeClaimNames.includes(vol.name));
```

src/stateful-set.ts
```diff
@@ -178,21 +208,76 @@ export class StatefulSet extends workload.Workload implements IScalable {
...
+
+      // If there are no volumes after filtering, don't include volumes property
+      if (filteredVolumes.length === 0) {
+        podSpec.volumes = undefined;
+      } else {
+        podSpec.volumes = filteredVolumes;
+      }
```
iliapolo, line 250: Will this cause an error?

```suggestion
        podSpec.volumes = filteredVolumes;
```

Worst case it just creates an empty volumes array, seem harmless.

iliapolo (changes requested):

PR #6126, opened 2025-08-18
feat: cdk8s-plus-33

iliapolo:
Following instructions: https://github.com/cdk8s-team/cdk8s-plus/blob/k8s-29/main/rotate.md#two-create-the-new-cdk8s-plus-branch


iliapolo:
Meh, wrong branch name.

PR #6382, opened 2025-10-11
feat(stateful-set): allow configuring `volumeClaimTemplates`

### LINKED ISSUE #2183: Support PersistentVolumeClaimTemplate in STS

benkeil:
Fixes cdk8s-team/cdk8s#2183


docs/plus/volume.md
```diff
@@ -36,3 +36,39 @@ const redis = pod.addContainer({
...
+import * as kplus from 'cdk8s-plus-32';
+import { Size } from 'cdk8s';
+
+// Create a StatefulSet with a PVC template
+const statefulSet = new kplus.StatefulSet(this, 'StatefulSet', {
+  containers: [{
+    image: 'nginx',
+    volumeMounts: [{
+      path: '/data',
+      name: Volume.fromName(chart, "pvc-template", "data-volume"),
```
iliapolo: To provide a more idiomatic example, extract this to a constant:

```ts
const volume = Volume.fromName(chart, "pvc-template", "data-volume");
```

Then use `volume.name` below in the template configuration.

docs/plus/volume.md
```diff
@@ -36,3 +36,39 @@ const redis = pod.addContainer({
 // mount to the redis container.
 redis.mount('/var/lib/redis', data);
 ```
+
+## Using PersistentVolumeClaim Templates with StatefulSets
```
iliapolo, line 40: Ha, we don't have specific docs for stateful sets. Note to self.

src/container.ts
```diff
@@ -212,7 +212,7 @@ export interface SeccompProfile {
   readonly type: SeccompProfileType;
 }
 
-export interface ContainerSecutiryContextCapabilities {
+export interface ContainerSecurityContextCapabilities {
```
iliapolo: Unfortunately we can't do this as its a breaking change :(

.gitignore
```diff
@@ -78,3 +78,4 @@ docs/typescript.md
 docs/python.md
 docs/java.md
 !/.projenrc.ts
+.idea
```
iliapolo: You need to add this via the `.projenrc.ts` file, otherwise it gets overridden during build.

src/stateful-set.ts
```diff
@@ -138,10 +152,10 @@ export class StatefulSet extends workload.Workload implements IScalable {
     this.apiObject.addDependency(this.service);
 
     this.replicas = props.replicas;
-    this.strategy = props.strategy ?? StatefulSetUpdateStrategy.rollingUpdate(),
+    this.strategy = props.strategy ?? StatefulSetUpdateStrategy.rollingUpdate();
     this.podManagementPolicy = props.podManagementPolicy ?? PodManagementPolicy.ORDERED_READY;
     this.minReady = props.minReady ?? Duration.seconds(0);
-
+    this.volumeClaimTemplates = props.volumeClaimTemplates;
```
iliapolo: call `addVolumeClaimTemplate` here so we benefit from its validations.

src/stateful-set.ts
```diff
@@ -279,3 +357,16 @@ export class StatefulSetUpdateStrategy {
...
+ * A PersistentVolumeClaim template for StatefulSets
+ */
+export interface PersistentVolumeClaimTemplateProps extends PersistentVolumeClaimProps {
+  /**
+   * The name of the claim that the StatefulSet controller will create for each pod.
+   * This will be used to name the created PVC in the format <claim-name>-<pod-name>
+   *
+   * This name should match the name of a volume mount in one of the containers.
+   */
+  readonly name: string;
```
iliapolo, line 378: Can we accept a `Volume` here instead? or a `VolumeMount`?

src/stateful-set.ts
```diff
@@ -279,3 +357,16 @@ export class StatefulSetUpdateStrategy {
...
+
+/**
+ * A PersistentVolumeClaim template for StatefulSets
+ */
+export interface PersistentVolumeClaimTemplateProps extends PersistentVolumeClaimProps {
+  /**
+   * The name of the claim that the StatefulSet controller will create for each pod.
+   * This will be used to name the created PVC in the format <claim-name>-<pod-name>
+   *
+   * This name should match the name of a volume mount in one of the containers.
```
iliapolo, line 376: This should be validated then.

src/stateful-set.ts
```diff
@@ -150,11 +164,25 @@ export class StatefulSet extends workload.Workload implements IScalable {
...
 
-  private _createHeadlessService() {
+  public addVolumeClaimTemplate(template: PersistentVolumeClaimTemplateProps): StatefulSet {
+    if (this.volumeClaimTemplates?.some(t => t.name === template.name)) {
+      throw new Error(`A volume claim template with name "${template.name}" already exists`);
+    }
+    this.volumeClaimTemplates = this.volumeClaimTemplates
+      ? [...this.volumeClaimTemplates, template]
+      : [template];
+    return this;
```
iliapolo: This isn't consistent with our APIs. Lets return void.

iliapolo (changes requested):

src/stateful-set.ts
benkeil: maybe change for all such functions in the future. would be no breaking change.

src/stateful-set.ts
benkeil: I would let it as it is. You could still use the volume as a variable and use `name: volume.name`, I would not use it like this.

I could also think about something like:

```ts
VolumeClaimTemplatesSomeSuffix.fromVolume(volume, {
  storage: Size.gibibytes(20),
  accessModes: [kplus.PersistentVolumeAccessMode.READ_WRITE_ONCE_POD],
  storageClassName: 'standard',
})
```

But to be honest my absolute favourite would be something like:

```ts
new kplus.StatefulSet(chart, 'StatefulSet', {
  containers: [
    {
      image: 'foobar',
      portNumber: 80,
      volumeMounts: [
        {
          volume: Volume.fromClaimTemplate({
            name: "data",
            storage: Size.gibibytes(20),
            accessModes: [kplus.PersistentVolumeAccessMode.READ_WRITE_ONCE_POD],
            storageClassName: 'standard',
          }),
          path: '/mnt/data',
        },
      ],
    },
  ],
})
```

docs/plus/volume.md
benkeil: updated

src/container.ts
benkeil: reverted.

.gitignore
benkeil: added

src/stateful-set.ts
benkeil: updated

src/stateful-set.ts
benkeil: added

src/stateful-set.ts
benkeil: removed

src/stateful-set.ts
iliapolo: I like `Volume.fromClaimTemplate` - probably a bit more involved to implement. Don't want to block on it so i'm approving this one. Feel free to add it in a subsequent PR though as an enhancement.

iliapolo (approved):

iliapolo:
@benkeil looks good, you got some build failures though.

iliapolo:
@benkeil new failures :)

benkeil:
wait until it runs in my fork

benkeil:
ok, build was successful

PR #6750, opened 2026-01-01
fix(stateful-set): restore missing volumes in stateful sets.

### LINKED ISSUE #6735: With StatefulSet, no volumes are added to the pod if there are no volumeClaimTemplates
I am building some custom constructs to deploy a service and I have noticed (while building it bit by bit), that my Volumes were not getting created when I was creating StatefulSet, but they did when I attempted the same with the deployment.

After bit more playing around, I found this code as a suspect of the issue. After adding a VolumeClaimTemplate, my non VolumeClaimTemplates have started appearing as expected.

https://github.com/cdk8s-team/cdk8s-plus/blob/6b50fe2f1ff18f6210520aa4c2bcf49a07f32f17/src/stateful-set.ts#L240-L254

I do believe that simplifying the method might remove he issue

```ts
private _filterPodSpecVolumes(podSpec: PodSpec) { 
   // When using volumeClaimTemplates, the volumes with matching names should not be included in the pod spec 
   const volumeClaimNames = (this.volumeClaimTemplates ?? []).map(vct => vct.name); 
   const volumesWithNoTemplates = (podSpec.volumes ?? []).filter(vol => !volumeClaimNames.includes(vol.name)); 

   // If there are no volumes after filtering, don't include volumes property 
   if (volumesWithNoTemplates.length === 0) { 
     return undefined; 
   } else { 
     return volumesWithNoTemplates; 
   } 
 } 
```

zaninime:
Fixes #6735.

### DIFF
--- a/src/stateful-set.ts
+++ b/src/stateful-set.ts
@@ -238,19 +238,14 @@ export class StatefulSet extends workload.Workload implements IScalable {
    * @internal
    */
   private _filterPodSpecVolumes(podSpec: PodSpec) {
-    // When using volumeClaimTemplates, the volumes with matching names should not be included in the pod spec
-    if (this.volumeClaimTemplates && podSpec.volumes) {
-      const volumeClaimNames = this.volumeClaimTemplates.map(vct => vct.name);
-      const volumesWithNoTemplates = podSpec.volumes.filter(vol => !volumeClaimNames.includes(vol.name));
-
-      // If there are no volumes after filtering, don't include volumes property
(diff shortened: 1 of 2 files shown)


zaninime:
Hello, can I get a review from anyone in @cdk8s-team? This is a small issue that prevents upgrades.

eminugurkenar:
this would be good to merge

patrickmscott:
I also just encountered this bug. Is this project still active? Not much activity for a while...

PR #7199, opened 2026-03-31
fix(stateful-set): preserve pod spec volumes when no volumeClaimTemplates are configured

### LINKED ISSUE #2858: cdk8s-plus stateful_set.ts _filterPodSpecVolumes() deletes all volumes if there are no persistent templates
The last line of `_filterPodSpecVolumes()` should return the full list of volumes as the default case.

https://github.com/cdk8s-team/cdk8s-plus/blob/c14b130aa43958568a7de67f0ef31b053b6da602/src/stateful-set.ts#L253

Currently if you have no volumeClaimTemplates configured it will just delete all your volumes.

officialasishkumar:
Fixes https://github.com/cdk8s-team/cdk8s/issues/2858

## Summary

The `_filterPodSpecVolumes()` method in `StatefulSet` returns `undefined` as the default case when no `volumeClaimTemplates` are configured. This causes all volumes (e.g., secret volumes, configmap volumes) to be silently dropped from the synthesized pod spec.

The fix changes line 253 in `src/stateful-set.ts` from `return undefined` to `return podSpec.volumes`, so that volumes are correctly preserved when `volumeClaimTemplates` are not in use.

## Changes

- **`src/stateful-set.ts`**: Return `podSpec.volumes` instead of `undefined` in the default branch of `_filterPodSpecVolumes()`
- **`test/statefulset.test.ts`**: Added a test case that creates a `StatefulSet` with a secret volume but no `volumeClaimTemplates`, and verifies the volume is present in the synthesized output

## Test plan

- [x] All 492 existing tests pass
- [x] New test case validates the fix
- [x] `npx projen compile` succeeds


zaninime:
Duplicate of #6750, also you basically copy-pasted the test I added.

